PR #1, opened 2021-01-18
Trigger on any change

GuyHarperDay:


digi-brain:
@GuyHarperDay I don't remember this PR, but it seems like whatever we were trying to do here we have since implemented anyway (plus, the path in this code is outdated now). Is it safe for me to close, or am I missing something?

PR #2, opened 2021-05-23
Add a defined Start phase

digi-brain:
Right, first attempt at pull request, I've added you as a 'reviewer' which seems logical. Let's see how it works.
You should be able to comment, both on individual lines and at higher levels. I'm not sure if you can edit directly in a PR review, but I imagine that you can do that too where it is the most expedient thing to do.
As it stands, I don't think you'll be able to merge changes onto the master branch as I'm the only one who has those permissions at present and I'm trying to work out the best workflows. But we can go back and forth until we're happy and then I'll merge into master when we're ready.


digi-brain:
One thing I forgot to do in this pass was to add notes as appropriate to indicate differences from the original E40k rules. Should probably add these on another pass before merge into the master branch.

IJW-Wartrader:
I've fixed a few 'aircraft' references that should be 'flyers', and added a 'prepare Initiative counters' step to the Start phase - both in the Start phase doc and in the turn sequence doc.

IJW-Wartrader (changes requested):
I think that's done, but please check my addition of a 'Prepare Initiative counters' step to the Start phase.

modules/main-rules/pages/the-start-phase.adoc
```diff
@@ -6,17 +6,21 @@ The Start phase comprises preparatory steps which aren't always necessary -- the
...
-* Some types of Special orders, such as Flak, Preparatory Bombardment and Dodge.
+* Flyer missions.
+* Some types of special orders, such as Flak, Preparatory Bombardment and Dodge.
 
 Start with the player whose army has the lower Strategy Rating, and then alternate to declare until you both run out of things to declare.
 Take Leadership tests in the normal way, where the rules require it for you to successfully set the orders you declare.
 
+== 3. Prepare Initiative counters
+
+The five Initiative counters are gathered together and placed in a mug or other receptacle.
```
digi-brain, line 21: All good here, except I'll rewrite nasty old passive case into active. :)

digi-brain:
@IJW-Wartrader I couldn't work out how to 'accept' your edits, but I agree with them all - I made further minor tweaks to just a couple of them. I'll send this to you to review again - let's see how it works. I do still need to add notes to highlight differences from the original E40k rules, but lets see how the next steps work and I'll do this as a separate task if necessary.

modules/main-rules/pages/the-start-phase.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+There are some things that you must declare now if you want to use them:
+
+* Flyer missions.
+* Some types of special orders, such as Flak, Preparatory Bombardment and Dodge.
+
+Start with the player whose army has the lower Strategy Rating, and then alternate to declare until you both run out of things to declare.
+Take Leadership tests in the normal way, where the rules require it for you to successfully set the orders you declare.
+
+== 3. Prepare Initiative counters
```
IJW-Wartrader, line 19:
```suggestion
== 3. Prepare the Initiative counters
```

OK, I think this should give you a suggestion that can be accepted. I've added 'the' to match the text in the main turn sequence page.

IJW-Wartrader: Yay, it's showing me a commit button, so you should be seeing it too.

digi-brain: Yes, I see it, yay! And as long as I go to the **Files** tab first, I can **Add suggestion to batch** so that we get one commit rather than an excess of little ones — that's good.

digi-brain (commented):
I can't 'approve' my own pull request, but I wonder if this 'comment' will clear the 'Changes requested' problem.

digi-brain:
Hmmm, nope, still blocked. I think I'll force the merge and hope that things become more clear next time.

PR #8, opened 2021-06-05
Special abilities PR 5, Split Speed values

IJW-Wartrader:
In the navigation, Cavalry has been moved to Data sheet modifiers.

Jump Packs hasn't been moved as we need to decide whether to split it into two pages or add a stub page in the list of Data sheet modifiers, or use a partial to put the same content in two pages.

I've added the same edition note to Data sheets and characteristics and Assault moves and charges, presumably this can be replaced with a partial.

In Assault moves and charges I've simply added the new bullet point for split Speed values, without attempting to match the phrasing of the current page. Do you want to think about the charge rewrite now, or leave it for a future branch?

**Style guide issues - units**

The style guide A-Z says to separate numerals and units with a space. I've done this in all the split Speed texts, but the existing texts like the samples on Data sheets and characteristics don't use a space. The original GW texts didn't either. Which do you prefer to use, going forwards?


modules/special-rules/pages/dodge.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+---
+
+Where your detachment has the {dodge} ability, you can use special orders called Dodge:
+
+. If you wish to place any detachments on Dodge orders then you must do this in the Start phase.
+. Place the {special-orders} marker to indicate that the detachment is on Dodge orders.
+While Dodge orders are in effect:
+ ** Increase the Armour value of all units in the detachment by one, to a maximum of 6+.
+ ** Halve the total Firepower of the detachment detachment.
```
digi-brain, line 20:
```suggestion
 ** Halve the total Firepower of the detachment.
```

modules/special-rules/pages/armour-1.adoc
```diff
@@ -0,0 +1,9 @@
+= Armour +1
+
+These units have particularly heavy armour for their kind.
```
digi-brain, line 3:
```suggestion
These units have particularly effective armour for their kind.
```

modules/special-rules/pages/brood-telepathy.adoc
```diff
@@ -5,6 +5,10 @@ This gives the Cult strong morale.
 
 ---
 
-While at least one Genestealer Patriarch or Magus is present in your Genestealer Cult detachment you automatically pass any Leadership test that you are required to make for them.
+Apply these rules to detachments with the {brood-telepathy} ability:
```
digi-brain, line 8:
```suggestion
Where your detachment has the {brood-telepathy} ability:
```

modules/special-rules/pages/data-sheet-modifiers.adoc
```diff
@@ -0,0 +1,9 @@
+= Data sheet modifiers
+
+These abilities to not provide special rules effects.
```
digi-brain, line 3:
```suggestion
These abilities do not provide special rules effects.
```

modules/special-rules/pages/data-sheet-modifiers.adoc
```diff
@@ -0,0 +1,9 @@
...
+Rather, they modify the characteristics in the unit's data sheet.
```
digi-brain, line 4:
```suggestion
Rather, they modify the characteristic values on unit data sheets.
```

modules/special-rules/pages/dauntless.adoc
```diff
@@ -0,0 +1,17 @@
+= Dauntless
+
+Some troops are so disciplined or relentless that they never flinch in the face of the enemy.
+
+[NOTE.e40k]
+====
+In comparison to _Epic 40,000_, we've separated this ability from parts of the Dark Angels Deathwing and Tyranid Synapse Creatures rules.
```
digi-brain, line 7:
```suggestion
In comparison to _Epic 40,000_, we derived this ability from parts of the Dark Angels Deathwing and Tyranid Synapse Creatures rules.
```

modules/special-rules/pages/firepower-1.adoc
```diff
@@ -0,0 +1,9 @@
+= Firepower +1
+
+These units are armed with potent weaponry that does not affect their range.
```
digi-brain, line 3:
```suggestion
These units are especially potent at range for their kind.
```

modules/special-rules/pages/hitch-x.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+---
+
+Where your unit has the {hitch} ability:
+
+* The {hitch} unit can carry as many Infantry units as we state in brackets.
+For example, the Battlefortress has {hitch-4} -- so you can embark up to four Infantry units, as we explain in the main transport rules.
+* The {hitch} unit does not gain the abilities of embarked units. For example a {hitch} unit carrying a unit with the {save} ability does not gain the {save} ability.
+
+NOTE: A {hitch} unit does not gain the abilities of embarked units.
```
digi-brain, line 13: Remove the redundant line, as the note that follows covers the same point and works better than a bullet point in this case.
```suggestion

NOTE: A {hitch} unit does not gain the abilities of embarked units.
```
We could add the example to the note, but I really don't think that such an example is necessary -- the note seems very clear and unambiguous to me.
If we did though, then I'd like to reword a little: "For example, where a unit that has the {save} ability is embarked on a {hitch} unit, the {hitch} unit does not gain the {save} ability."

modules/special-rules/pages/infiltrators.adoc
```diff
@@ -1,2 +1,15 @@
 = Infiltrators
 
+These units are the expert scouts and commandos who gather information, set ambushes, sabotage installations, and so on.
+
+---
+
+Apply these rules to units with the {infiltrators} ability:
```
digi-brain, line 7:
```suggestion
Where your unit has the {infiltrators} ability:
```

modules/special-rules/pages/infiltrators.adoc
```diff
@@ -1,2 +1,15 @@
...
+
+* Count them as *in command* while they are within 60cm of their HQ unit, rather than the usual 30cm.
+However, they are still destroyed if they become Broken and they are more than 30cm away from their HQ unit.
+* You may deploy them up to 30cm outside of your normal deployment zone for the scenario you are playing.
```
digi-brain, line 11:
```suggestion
* You may deploy them up to 30cm outside of your normal deployment zone for the scenario that you are playing.
```

modules/special-rules/pages/synapse.adoc
```diff
@@ -0,0 +1,15 @@
+= Synapse
+
+Where Synapse creatures are present, the Tyranid Hive Mind is able to impose maximum control over its swarms. 
+
+---
+
+Where your detachment has the {synapse} ability:
+
+* While the detachment includes any {synapse-creature} units, apply the {dauntless} and {unbreakable} abilities.
++
```
digi-brain, line 10: Delete redundant line with rogue `+`.
```suggestion
```

modules/special-rules/pages/walker.adoc
```diff
@@ -0,0 +1,22 @@
+= Walker
+
+Some vehicles with legs are particularly manoeuvrable.
```
digi-brain, line 3:
```suggestion
Some vehicles that have legs are able to enter terrain that is problematic for other vehicles.
Thus they can fill niche roles on the battlefield, often in close support of infantry.
```

digi-brain (commented):
Some enhancements to wording - I might go ahead and commit some of these where they seem to be 'no brainers' so we don't have too many unresolved conversations hanging around.
I'll get a close look at the more substantive rules text as soon as I can.

modules/special-rules/pages/transport-x.adoc
```diff
@@ -1,2 +1,16 @@
...
+
+---
+
+Where your unit has the {transport} ability:
+
+* The {transport} unit can carry as many Infantry units as we state in brackets.
+For example, the Rhino has {transport-2} -- so you can embark up to two Infantry units in a Rhino, as we explain in the main transport rules.
+
+NOTE: A {transport} unit does not gain the abilities of embarked units.
+For example a {transport} unit that carries a unit with the {save} ability does not gain the {save} ability.
```
digi-brain, line 13: As with hitch-x.adoc - I don't think we really need the example.
```suggestion
```
Though if we do keep it then I'd like to reword in similar fashion to my comment at hitch-x.adoc

modules/main-rules/pages/transport.adoc
```diff
@@ -5,35 +5,31 @@ We explain special abilities in more detail at xref:special-rules:transport-x.ad
 
 Units that have the {transport} special ability can carry other units across the battlefield.
 
-* To embark or disembark other units (or cargo) it costs the {transport} unit 5cm of its movement.
+* To embark or disembark other units (or cargo) costs the {transport} unit 5cm of its movement.
 * An embarking or disembarking unit must also pay 5cm of its movement.
 * You may embark and disembark units in the same turn, the same phase or even in the same move -- so long as all of the involved units have enough movement.
-* You may perform no other actions with units that are embarked -- so, you cannot shoot with them, for example. 
+* You may perform no other actions with units that are embarked -- so, you cannot shoot with them, for example.
 Obviously, once you disembark them then the normal rules apply again.
```
IJW-Wartrader, line 12: I just remembered some discussions about how to measure movement when disembarking, and whether you should place the unit then measure, or measure and then place the unit.

The main note is a bit wordy, let me know if you can think of something better.

```suggestion
Obviously, once you disembark them then the normal rules apply again.

NOTE: Measure the movement of disembarking units from the base of the {transport} unit. If there is not enough movement to clear the {transport} unit's base, place the disembarking unit in base contact with the {transport} unit.

[NOTE.e40k]
====
This was not specified in _Epic 40,000_, but reduces the effects of different size bases.
====
```

modules/main-rules/pages/data-sheets-and-characteristics.adoc
```diff
@@ -57,7 +57,14 @@ Some rules affect some unit types and not others.
 
 Speed:: The distance that you can move the unit in a single movement action (in centimetres).
 It represents a rate at which the unit can move across the battlefield and still be effective when it shoot its weapons.
-You can give units orders that allow you to move units farther than this, but their shooting will suffer.
+You can give units orders that allow you to move units farther than this, but their shooting will suffer. +
++
+Where a unit has a split Speed value such as 20/30 cm, use the first value for all purposes except Assault moves.
```
IJW-Wartrader, line 62: As per the Issue discussion.

```suggestion
Where a unit has a split Speed value such as 20/30 cm, use the first value for all purposes except charging into base contact.
```

digi-brain:
> In the navigation, Cavalry has been moved to Data sheet modifiers.
> 
> Jump Packs hasn't been moved as we need to decide whether to split it into two pages or add a stub page in the list of Data sheet modifiers, or use a partial to put the same content in two pages.
> 
> I've added the same edition note to Data sheets and characteristics and Assault moves and charges, presumably this can be replaced with a partial.
> 
> In Assault moves and charges I've simply added the new bullet point for split Speed values, without attempting to match the phrasing of the current page. Do you want to think about the charge rewrite now, or leave it for a future branch?
> 
> **Style guide issues - units**
> 
> The style guide A-Z says to separate numerals and units with a space. I've done this in all the split Speed texts, but the existing texts like the samples on Data sheets and characteristics don't use a space. The original GW texts didn't either. Which do you prefer to use, going forwards?

I've only just noticed this. My thoughts:

Partial: It's a bit of a toss up sometimes as to whether or not it is really worth the bother when you only have two instances of something fairly simple. I suggest leave it as-is for now, and when I'm checking the wordings and structure of stuff I'll create a partial if I think it is worthwhile on balance. 

Charges: I'd prefer to keep with existing charge move terminology for now, and then experiment with a rewrite as a separate issue with a branch for this purpose. Let's see how well a rewrite works independently of the rest, and not extend the scope of current work in progress if we can help it - so yes, separate issue and branch is best I think.

Style guide, separate numerals and units with a space: I compiled the style guide relatively recently, so some stuff I'd written earlier doesn't conform and I'll want to improve that in due course. Space between the number and the units is standard practice in good style guides, for a couple of reasons. Some exceptions are inevitable, given the complexity of language of course. Eg, from Home Office Science style guide:

> There is a space between a number and a unit of measurement e.g. 20 mg – except for temperatures e.g. 20°C and percentages e.g. 20% (or in text use 20 per cent). When giving a range do not repeat the unit, e.g. 5 to 8 mg and not 5 mg to 8 mg. Use the word 'to' in text, not a dash.

It matters a lot less in this project, because we really only use 'cm' units and don't expect any other elements where lack of a space can be a problem as it can be in engineering and science writing. But still, might as well follow the common standard.

Arguably, in data sheets we should put the units in the Speed/Range column headers ["Speed (cm)"/"Range (cm)"] rather than in every cell. But perhaps having it in every cell provides a more convenient visual cue. I could be persuaded either way on that one.

IJW-Wartrader:
> Charges: I'd prefer to keep with existing charge move terminology for now, and then experiment with a rewrite as a separate issue with a branch for this purpose. Let's see how well a rewrite works independently of the rest, and not extend the scope of current work in progress if we can help it - so yes, separate issue and branch is best I think.

Aye. I'll tweak the Assault moves and charges addition to match.

> Arguably, in data sheets we should put the units in the Speed/Range column headers ["Speed (cm)"/"Range (cm)"] rather than in every cell. But perhaps having it in every cell provides a more convenient visual cue. I could be persuaded either way on that one.

I'd prefer to keep the unit. Partly for familiarity because that's what I'm used to seeing for E40k units, and partly because it reinforces that Speed and Range aren't pure numerical values in the way that Assault and firepower are.

digi-brain:
> I'd prefer to keep the unit. 
Fine with me. As I say, I could've been persuaded the other way, but to date I kept the units in the cells for the cues and consistency with what went before, and I'm happy to continue like that.

modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -26,4 +26,10 @@ This is the only time that you may move your units into base contact with enemy
 * You don't have to declare and make a charge into close combat unless you want to, with one exception: If you can only meet the 'must move towards the enemy' requirement by moving toward an enemy that has units that are within 5cm of your units, then you must charge and place units in base contact.
 * For a charge, you may double the Speed value of *infantry* units *if* this will bring them into base contact with enemy units.
 * If you don't get at least one unit into base contact with an enemy then there is no charge and there will be no close combat.
-(Though there will be a firefight if any units are within 15cm of the enemy.)
\ No newline at end of file
+(Though there will be a firefight if any units are within 15cm of the enemy.)
+* Where a unit has a split Speed value such as 20/30 cm, you may use the second value if this would bring the unit into contact with an enemy. You *cannot* double the Speed value of the unit during an Assault move. +
```
IJW-Wartrader, line 30: Rewrite of the split Speed text to fit better.

```suggestion
* For units with a split Speed value such as 20/30 cm, you may use the second value for a charge *if* this will bring them into base contact with enemy units. This is *instead* of doubling the units' Speed value. +
```

modules/main-rules/pages/data-sheets-and-characteristics.adoc
digi-brain: I think that this is all too much detail for this topic. Remember, this is one of the 'Basics' topics, and while we write in 'every page is page one' topics, some level of 'progressive disclosure' still has value and I think that it is reasonable to assume that newbies will tend to read this long before they get to Assault rules and charge moves and such.

I do think that it is good to acknowledge the split speeds here though, and to reference the section that explains them more. So, I'll edit with this in mind and push an update later then see what you think.

modules/main-rules/pages/transport.adoc
digi-brain: I don't 100% understand what you're trying to say here, so may be good to talk it through. I'm sure we've discussed it before, but I've forgotten the crux of it.  I intend to make some other edits here anyway, so I'll push those later and then we can hone it some more to get across the exact meaning you intend in the best way.

modules/main-rules/pages/data-sheets-and-characteristics.adoc
IJW-Wartrader: > I do think that it is good to acknowledge the split speeds here though, and to reference the section that explains them more. So, I'll edit with this in mind and push an update later then see what you think.

What section are you referring to? The Assault moves section obviously details how the second value works, but you can't move the general explanation of a characteristic there.

digi-brain:
Among other things, I've pushed some changes to `assault-moves-and-charges.adoc`. I do think that whether or not we keep the 'charge' terminology in the end, this topic will need further changes to make it easy to understand. It's slightly tricky to clarify the general idea of the Assault move towards the enemy, a detachment becoming engaged in close combat as a whole ('charge') and making charge moves/base contact with individual units.

modules/main-rules/pages/transport.adoc
IJW-Wartrader: The crux of it is that if you place a disembarking unit in base contact with the Transport, and **then** measure the disembarking unit's movement, different base sizes have a substantial effect on how far the unit can move. 

For example a unit on a 20mm square base unit will lose a couple of centimetres of movement compared to a unit on a 40x12.5mm strip base.

digi-brain: I'll resolve this conversation as I've used this suggestion as the core for changes which now appear below.

modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -4,26 +4,45 @@ During the Assault phase you may only move and attack with detachments that you
...
-* For a charge, you may double the Speed value of *infantry* units *if* this will bring them into base contact with enemy units.
-* If you don't get at least one unit into base contact with an enemy then there is no charge and there will be no close combat.
-(Though there will be a firefight if any units are within 15cm of the enemy.)
\ No newline at end of file
+You can benefit from a special increase to the Speed value of some particular units when you make a charge move with them.
+(Remember, it is only a charge move if the units will get into base contact with the enemy -- you don't get the increase if they won't get into base contact.)
+
+* You may double the Speed value of any *Infantry* unit.
+* Where a unit has a split Speed value such as `20/30 cm`, you may *not* double the first value -- use the second value instead.
+* For any other unit, must use their Speed value in the usual way.
```
IJW-Wartrader, line 38: This is ambiguous, and can be read as doubling the second value instead of doubling the first value.

I've also removed an errant 'must' from the next bullet.

```suggestion
* Where a unit has a split Speed value such as `20/30 cm`, use the second value *instead* of doubling the first value.
* For any other unit, use their Speed value in the usual way.
```

modules/main-rules/pages/data-sheets-and-characteristics.adoc
```diff
@@ -57,7 +57,10 @@ Some rules affect some unit types and not others.
 
 Speed:: The distance that you can move the unit in a single movement action (in centimetres).
 It represents a rate at which the unit can move across the battlefield and still be effective when it shoot its weapons.
-You can give units orders that allow you to move units farther than this, but their shooting will suffer.
+You can give units orders that allow you to move units farther than this, but their shooting will suffer. +
++
+Some unit data sheets show a split Speed value such as `20/30 cm`.
+For more about this, see xref:assault-moves-and-charges.adoc[].
```
IJW-Wartrader, line 63: As I said before, **THIS DOES NOT WORK**. 

Players need to know which value to use as the default, and that can't be buried in the Assault moves page. Not that you've explained it there anyway.

If you really dislike 'use the first value for all purposes except Assault moves' as being too specific, we can change it to 'use the first value unless otherwise specified' or similar, but I don't see the benefit if you're going to reference Assault moves anyway.

digi-brain: Okay, no need to shout, and I'm not sure what exactly you've said before about this that makes you feel so outraged.

I don't see what the problem is. It's a basic, introductory topic — the main rules are built with progressive disclosure in mind, so they try not to overwhelm the reader with all the whys and wherefores of stuff that comes from special rules, as Split speed values do. (One of the reasons I was trying to think through another approach in that other conversation.) This is normal.

It referenced topic **does** explain it, based on the text that you yourself added but with some edits:

> * Where a unit has a split Speed value such as `20/30 cm`, you may *not* double the first value -- use the second value instead.

However, I can see that we've not explicitly mentioned that the player should use the first value for everything else. That is a valid thing to fix, but again, no shouty shouty needed. If you want things to be perfect in a single pass, in terms of rules and language and structure, with no discussion or iteration required, then you're on a fools errand.

How about something like this?
```suggestion
Some unit data sheets show a split Speed value such as `20/30 cm`.
Use the first value for all movement, except where the rules tell you otherwise.
For information about the second value, see xref:assault-moves-and-charges.adoc[].
```

modules/main-rules/pages/assault-moves-and-charges.adoc
digi-brain: Well spotted, I agree that it could be read that way, and indeed there is an errant 'must'.
Thinking about it, I don't think we really need to mention "don't double" because that only applies to the previous bullet anyway — strictly, we don't use the second value instead of double-Speed (which we wouldn't do anyway for Cavalry because they count as Vehicle not Infantry). Rather, we use the second value instead of the normal Speed (first value), as one of a couple of special cases for a charge move .
```suggestion
* Where a unit has a split Speed value such as `20/30 cm`, use the second value.
* For any other unit, use their Speed value in the usual way.
```

modules/main-rules/pages/transport.adoc
digi-brain: >The crux of it is ...

Lol, I think you replied before I posted my message to say I would mark resolved because the new content is below. I'll unresolve so I have a prompt to look at it again. 

Gotcha, I remember now. I'll wordsmith a little more to make this more explicit.

modules/main-rules/pages/assault-moves-and-charges.adoc
digi-brain: Will resolve conversation, as outdate - newer commit has fixed this.

modules/main-rules/pages/data-sheets-and-characteristics.adoc
digi-brain: Outdated, fixed in a newer commit. So I'll resolve this conversation.

modules/main-rules/pages/assault-moves-and-charges.adoc
IJW-Wartrader: Don't forget that this covers Jump Packs units as well, which as Infantry **do** default to getting double Speed. 

Maybe:

```suggestion
* Where a unit has a split Speed value such as `20/30 cm`, use the second value but you cannot double it.
* For any other unit, use their Speed value in the usual way.
```

modules/special-rules/pages/regroup.adoc
```diff
@@ -0,0 +1,9 @@
+= Regroup
+
+Space Marines are known for their stubborn determination, and yet troops like the Dark Angels Deathwing have a special reputation for regrouping effectively, no matter the circumstances.
+
+---
+
+Apply these rules to detachments with the {regroup} ability:
```
IJW-Wartrader, line 7: Just spotted this older phrasing.

```suggestion
Where your detachment has the {regroup} ability:
```

modules/special-rules/pages/save.adoc
```diff
@@ -1,2 +1,14 @@
...
+These units can beat the odds and escape destruction -- perhaps through exceptional armour, protective fields, or extreme skills and toughness.
+
+---
+
+Where your unit has the {save} ability:
+
+* Each time the unit suffers a hit, from any source, roll a die.
+On a roll of 4+, ignore the hit -- it causes no damage.
+
+NOTE: Attacks which ignore void shields or power fields do *not* ignore the {save} ability.
```
IJW-Wartrader, line 12: Just spotted that there is already a bullet for this in holo-fields which we can borrow:

```suggestion
* The {save} ability applies to attacks that ignore void shields and power fields.
```

modules/main-rules/pages/data-sheets-and-characteristics.adoc
```diff
@@ -57,7 +57,11 @@ Some rules affect some unit types and not others.
 
 Speed:: The distance that you can move the unit in a single movement action (in centimetres).
 It represents a rate at which the unit can move across the battlefield and still be effective when it shoot its weapons.
-You can give units orders that allow you to move units farther than this, but their shooting will suffer.
+You can give units orders that allow you to move units farther than this, but their shooting will suffer. +
++
+Some unit data sheets show a split Speed value such as `20/30 cm`.
+Use the first value for all movement, except where the rules tell you otherwise.
```
IJW-Wartrader, line 63: This is much better, but 'all movement' is too restrictive which is why I used 'all purposes'.

Reserves rolls and Distortion Cannon both reference the unit's Speed value directly, without being a form of movement.

Unless you have another word or structure that you'd prefer to use instead of 'purpose':

```suggestion
Use the first value for all purposes, except where the rules tell you otherwise.
```

modules/main-rules/pages/transport.adoc
IJW-Wartrader: That looks good, resolving the conversation!

modules/special-rules/pages/regroup.adoc
digi-brain: Yeah, I've fixed a couple of those in other commits too. I'll go ahead and apply this one, and get a look at the rest of the text another time.

modules/main-rules/pages/assault-moves-and-charges.adoc
digi-brain: Balls. Of course, the preceding bullet doesn't work for Infantry that have Jump Packs. I'll add a new suggestion which covers all the lines together.

digi-brain, line 38:
```suggestion
* You may double the Speed value of any *Infantry* unit, unless it has a split Speed value.
* Where a unit has a split Speed value such as `20/30 cm`, use the second value.
* For any other unit, use its Speed value in the usual way.
```

digi-brain (commented):
Comments and suggestion to resolve 'split Speed value' issue in the bullets about special bonuses for charge moves.

modules/main-rules/pages/assault-moves-and-charges.adoc
digi-brain: I'll resolve this conversation, as we have a new conversation on all three bullet rather than just the latter two.

modules/main-rules/pages/transport.adoc
digi-brain: Yeah, I didn't get to further word-smithing yet to make that point more explicit. But now that you've reminded me of the crux of the point and I look at the word-smithing I already did, I agree that it is probably adequate and we can get away with not making it more explicit.

Though frankly I don't know why anyone would ever place the unit and then measure. To me, that is obviously wrong -- as wrong as measuring from the front of the base to the back where again you'd gain the width of the base.

Something further occurs to me. There appears to be no reason why you couldn't clear the transport unit's base - you must hold back enough movement to do it, on top of the 5cm cost for embarking/disembarking. We could make *that* explicit. But where the problem really lies IIRC is when you have *lots* of units such that potentially you can't fit them all around the transport (Orkeosaurus). In that case, it isn't clearing the base of the transport that is the problem, it's running out of space even when you've tried to do it all right, isn't it?

modules/main-rules/pages/data-sheets-and-characteristics.adoc
digi-brain: I see what you mean. One could argue that Reserves and the roll for Distortion cannon are still about movement — but it is indeed arguable and therefore ambiguous, and perhaps there are other less arguable edge cases.
I'm happy with that, I'll apply it and resolve.

modules/main-rules/pages/assault-moves-and-charges.adoc
IJW-Wartrader: Looks good to me.

digi-brain: Great, I'll commit it and resolve. This is nice, we're definitely honing things between us to be much more clear, complete and accurate than our first passes.

modules/special-rules/pages/save.adoc
IJW-Wartrader: This change looks fine, but Holo-fields now uses a completely different format from Save for a nearly-identical ability.

modules/special-rules/pages/holo-fields.adoc
```diff
@@ -0,0 +1,23 @@
...
+However, Eldar titans are of light construction -- they are built for speed and agility, not resilience.
+
+---
+
+Where your unit has the {holo-fields} ability:
+
+* Each time the unit suffers a hit, from any source, roll a die for a {holo-fields} save:
+ ** *2+*: Successful save, ignore the hit -- it causes no damage.
+ (But place Blast markers as normal.)
+ ** *1*: Unsuccessful save, apply the hit as normal.
```
IJW-Wartrader, line 15: Personally, I'm not a fan of this change. It feels very wordy without being clearer than the previous 'on a roll of X+, do Y' structure. 

Also, the 'saving roll' part of Save and Holo-fields is exactly the same ability with a different target number, so they ought to use the same structure.

IJW-Wartrader: The rest of the page looks better, though!

modules/main-rules/pages/transport.adoc
IJW-Wartrader: P.S. As far as the actual pull request is concerned, I think this can be marked as resolved.

modules/special-rules/pages/save.adoc
digi-brain: Looks like the exact same format to me -- it's just that Holo-fields has more clauses because it is more complex what with the automatic critical damage and all, and thus not a 'nearly-identical' ability.

In terms of process, I do agree that we want the presentation of similar things to be consistent — but the first priorities should be for each topic to be accurate and clear. So, it can be that when you're dipping in an out as you find time, and as you encounter new clauses and issues for which previous phrasing doesn't work well, it becomes inconsistent in these respects.

It's good to catch this as we go along where we can, but we shouldn't be **slaves** to it. That is, we can fix it in on a later iteration once we're able to step back and see the whole section -- this can be more efficient than constantly going go back and forth to tinker with it while fundamentals are still in flux during our first go at it.

So, good to note where we have inconsistencies, I agree -- but we shouldn't necessarily jump to fix them across the branch immediately every time.

modules/special-rules/pages/holo-fields.adoc
digi-brain: I'm open to other ideas, but keep in mind that in the previous iteration the term "unsaved" came out of nowhere, with no reference to 'saving'. Also you have to be careful in how you describe things like 'critical hit roll' and similar, because (a) the correct term is critical damage, and (b) it can be ambiguous as to whether you mean to roll to see if you get critical damage, or to roll to see what the critical damage result is.

So, it may be more wordy and have room for improvement, but I think that it **is** clearer for these reasons. I don't make these changes because I have some arbitrary preference that I want to impose -- I make them because I perceive a problem in the structure, logic and/or phrasing (and some other reasons too to do with best practices, consistency, etcetera).

modules/special-rules/pages/save.adoc
IJW-Wartrader: > Looks like the exact same format to me -- it's just that Holo-fields has more clauses because it is more complex what with the automatic critical damage and all, and thus not a 'nearly-identical' ability.

I think this was a case of crossed-wires. My comment about different formats was about the ability to ignore hits on a roll of X+, not about the note! See the Holo-fields conversation for more detail, and issues about Github not showing which lines I was referring to. :-(

> It's good to catch this as we go along where we can, but we shouldn't be **slaves** to it. That is, we can fix it in on a later iteration once we're able to step back and see the whole section -- this can be more efficient than constantly going go back and forth to tinker with it while fundamentals are still in flux during our first go at it.

Sure. It's just that, for me, Save and Holo-fields are variants of the same rule so I automatically think about both of them at the same time.

modules/special-rules/pages/holo-fields.adoc
IJW-Wartrader: Anyway, let's mark this as resolved. We can talk about which structure is a better fit at another time, and not let it hold up the pull request and branch merge.

I'd like to start on new content at some point!

digi-brain: Okay, let's mark resolved and move on. But so you have some idea where I'm coming from with these things:

Previously:

>* Each time the unit suffers a hit, from any source, roll a die. On a roll of 2+, the hit is ignored and causes no damage.
>* The {holo-fields} ability applies to attacks that ignore void shields and power fields.
>* Any unsaved hit on the unit is automatically a critical hit. Attacks which already automatically cause critical hits only cause one critical hit.
>
>NOTE: Saved hits still count as hits for placing Blast markers.

Yes, the text from Save would've worked to improve that first line, to be consistent and remove passive language etcetera --  **in isolation**. But it isn't in isolation.

Some of issues that the rest of the text poses:

1. 'Unsaved hit' (and then later, 'Saved hits') -- what's that? There is no prior mention of anything to do with save, saving, saved or similar. Yes, most players will know what we mean... but that's assumption, not explanation.
2. The *Any unsaved hit ..." bullet doesn't flow logically from the lead-in sentence. And even if it did, I think this one works better as a note, on the whole.
3. Given the restructure, we can eliminate the note for Blast markers and put it in the first new bullet, thus putting it where it is most relevant and minimising the mental 'overhead' of notes.
4. And other stuff probably, this is long enough.

The 'critical hit roll' thing is a general point, not an argument about what specific words you did or didn't use. I'm saying that the bullets tie in with the later paragraph to a degree, and one has to be careful in the later paragraph not to be ambiguous about critical hits and critical damage. Sometimes that gets a bit wordy. Perhaps it can be rationalised more, but it's not as straightforward as you're saying, and it all takes time so sometimes "clearer than it was" is good enough for the moment.

And yes, GW were not consistent about critical hits and critical damage. The actual thing is critical damage [War engine critical damage](https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/WarEngines/WarEngineCriticalDamage.htm). Potentially, if we defined it properly and were really consistent, 'critical hit' could mean 'a hit which also causes critical damage either automatically or because you rolled a 6 for it per the critical damage rules', maybe -- but that's a tangent to the main thrust here.

So, I'd like to get onto new content at some point too - but I'll always be behind you a bit because the bulk of my word-smithing logically must come after the bulk of your rules-lawyering, plus sometimes I can only find time to dip in and out. I've checked off most of the files in this PR, but still have some to do. And when I do, you can and should check that I haven't broken any important meaning -- but, so long as the meaning is correct, try not to focus overmuch on my particular choice of words or structure or this will be slow.

digi-brain: I got distracted a bit there, but what I'm trying to say in relation to the particular lines you clarified that you really meant to focus on, the Save text alone wouldn't connect well with the whole 'unsaved hits, critical damage' bit. Hence the bullets, to presage that. And in itself, this is not an uncommon device in GW rules either.
Anyway, we can change in future if we have a better way - I was just trying to explain why I change this sort of thing so we don't have to debate it every time.

modules/main-rules/pages/transport.adoc
digi-brain: Some interesting points, and definitely room for some more nuance here that we might want to explain eventually. But I agree, let's resolve this for now and revisit in future if and when we feel the urge.

modules/special-rules/pages/holo-fields.adoc
IJW-Wartrader: I think we're still talking at cross-purposes, but please mark it as resolved.

modules/special-rules/pages/save.adoc
digi-brain: >for me, Save and Holo-fields are variants of the same rule so I automatically think about both of them at the same time.
Yeah, that's fair enough, and logical. Ideally I look at 'similar' topics in parallel, and often do -- but sometimes I'm in a hurry or I forget because of where my focus is at that particular moment.

modules/special-rules/pages/cavalry.adoc
```diff
@@ -4,11 +4,33 @@ Despite the abundance and variety of technologies in the galaxy of the 41^st^ mi
...
+====
+In comparison to _Epic 40,000_, we've separated the Armour bonus part of the _Cavalry_ ability into our new {armour-1} ability.
+====
+
 ---
 
-Apply these rules to {cavalry} units, relative to the base units:
+Where your unit has the {cavalry} ability, apply these changes relative to its base unit:
+
+* Create a split Speed value.
```
IJW-Wartrader, line 19: A minor suggestion to cover a gap between creating the split value and applying it to the profile:

```suggestion
* Create and apply a split Speed value.
```

modules/special-rules/pages/jump-packs.adoc
```diff
@@ -1,2 +1,38 @@
 = Jump Packs
 
+These units are equipped with jet-powered devices or other technology, wings, or even powerful long legs that allow them to make long jumps across the battlefield.
+The resultant speed and mobility usually makes for good assault troops.
+
+---
+
+Where your unit has the {jump-packs} ability, apply this change relative to its base unit:
+
+* Create a split Speed value.
```
IJW-Wartrader, line 10: A minor suggestion to cover a gap between creating the split value and applying it to the profile:

```suggestion
* Create and apply a split Speed value.
```

modules/special-rules/pages/cavalry.adoc
digi-brain: Interesting. I'm not sure you can create it without also applying it, since to 'create' it you'd have to put it on the data sheet, so you've already applied it to the data sheet. :-D However, if you would argue this point and you think it reads better, I'm happy to go ahead with your suggestion here and on Jump Packs. Please feel free to apply them both in that case.

Though I suppose that now you've made me think of it, the thing that is really missing here is mention of the data sheet (it's implied by extension, but not explicit)... which could be in the bullet, or probably even better in the lead-in sentence. But then we should also make other similar things match. So, I'm happy to leave that for now too -- it's the type of thing that I think is better to refine across the whole section sometime in future.

modules/special-rules/pages/jump-packs.adoc
digi-brain: Ditto my comment at cavalry.adoc.

modules/special-rules/pages/cavalry.adoc
IJW-Wartrader: > Interesting. I'm not sure you can create it without also applying it, since to 'create' it you'd have to put it on the data sheet, so you've already applied it to the data sheet. :-D

That's exactly the problem - you've created a value, and then done nothing with the created value. Compare to other modifying ability texts like 'Add X to their Y value' or 'double their Y value', which inherently refer to the unit's data sheet.

In programming terms, you've created a variable and put a value in it, but that variable is 'invisible' because it's not referred to or used anywhere.

Hmm. Maybe like this?

```suggestion
* The unit has a split Speed value, created in this way.
```

Or

```suggestion
* The unit's Speed characteristic has two values.
```

IJW-Wartrader: I think the second one is closer to existing Data sheet modifier abilities.

modules/special-rules/pages/jump-packs.adoc
IJW-Wartrader: Copying the revised suggestions here for completeness.

```suggestion
* The unit has a split Speed value, created in this way.
```

Or

```suggestion
* The unit's Speed characteristic has two values.
```

modules/special-rules/pages/cavalry.adoc
digi-brain: >The unit has a split Speed value, created in this way.
>Or
>The unit's Speed characteristic has two values.

The first one still doesn't actually address the point -- the data sheet is still implicit. The second one does address the point because it references 'characteristic', which is on the data sheet.

However... the unit doesn't 'have' ('has') anything until you make it so. Both of these suggestions are in passive case -- so they don't fit with the active case, second person, imperative style of the surrounding lines and most of the rest of publication.

I could write a sentence in active case etc. that does reference the data sheet -- but as I mentioned, probably the best solution is to change the lead-in sentence (so that you don't have to repeat in each bullet where there are multiple). And then apply the same across all similar cases. So, could do it now, but I'm happy to leave it for the moment and fix in one swoop later, probably at the same time as other similar consistency stuff across the whole section.

As it is, I have only to review the transport topic and check it back with you and the whole PR is ready to merge.

digi-brain: Or actually, I think this would do it with no bother, in similar fashion to 'Add X to their Y value'...
```suggestion
* Make its Speed value into a split Speed value.
```

modules/special-rules/pages/jump-packs.adoc
IJW-Wartrader: Copying the most recent suggestion here so that it can be committed easily.

```suggestion
* Make its Speed value into a split Speed value.
```

modules/special-rules/pages/cavalry.adoc
IJW-Wartrader: > Or actually, I think this would do it with no bother, in similar fashion to 'Add X to their Y value'...

Let's go with that. I'll copy it across to Jump Packs and commit them both.

modules/main-rules/pages/transport.adoc
```diff
@@ -5,35 +5,44 @@ We explain special abilities in more detail at xref:special-rules:transport-x.ad
...
 
+[NOTE]
+.Placement of disembarked units
+====
+When you disembark units, measure their movement from the base of the {transport} unit.
+If you do not have enough movement to clear the {transport} unit's base then place the disembarked unit in base contact with the {transport} unit.
+====
```
digi-brain, line 19: I'm editing this topic at the moment and will share my efforts soon. But on these particular lines, I wonder as follows:

As written, might this permit cases where players could have sacrificed some movement, but don't because the rules say they can put the units in base contact anyway - so they use that to get a 'free' gain?

In general, I feel like perhaps there should be some concept of not exceeding the 'calculated maximum move you could get from the transport plus the embarked unit in combination' -- even if this means you must move transport unit a bit less or something to compensate rather than adding on the width of the disembarked unit's base and thus going beyond the calculated maximum move.

If there is something in this, I'd be happy for us to create an issue to tackle this properly later - just wanted to raise the point and get your thoughts.

IJW-Wartrader: > As written, might this permit cases where players could have sacrificed some movement, but don't because the rules say they can put the units in base contact anyway - so they use that to get a 'free' gain?

As far as I can see, this would require them to have a base longer than 5 cm. 

Don't forget that the unit is **already** sacrificing 5 cm of move just to be able to disembark. This is another reason why I automatically assumed that you 'deploy then measure' - as far as I was concerned, the unit had paid a movement penalty to get out, why would the unit have to pay a **second** movement penalty by measuring from the transport to 'move' out?

As a reminder, don't forget that what we currently have is already going to be regarded as a controversial rules change by some players, and if they've played 'deploy then measure' for a long time it's going to be unpopular, at best.

> In general, I feel like perhaps there should be some concept of not exceeding the 'calculated maximum move you could get from the transport plus the embarked unit in combination' -- even if this means you must move transport unit a bit less or something to compensate rather than adding on the width of the disembarked unit's base and thus going beyond the calculated maximum move.

My gut reaction is that this part of the map is marked 'Here be Dragons'. It's likely to be hard to write clear concise rules for, the overall subject is already controversial/generates lots of arguments, it's unlikely to get widespread support from the players (and that includes me), and it adds even more inconsistencies with Hitch. Or you apply it to Hitch as well, and screw over the Orks even more.

Paraphrasing...

'You know that unstoppable Battlefortress war engine that isn't slowed down at all when Da Boyz leap suicidally off the side?'
'Yer.'
'Well, it just slowed down by 4 cm because some of Da Boyz want to leap suicidally off the front instead of the side.'

P.S. In places where I'd normally use italics for emphasis I'm using bold on Github for consistency with the style guide.

modules/main-rules/pages/transport.adoc
```diff
@@ -1,39 +1,61 @@
-= Transport
+= Transportation
```
digi-brain, line 1: This would've always been a better title I think, and it helps to disambiguate from the ability (and from the flyer mission, though we need to improve those titles and filenames later anyway).

I should change the filename to match too - but I will also need to find and update all inbound links to use the new filename, so I'll see what you think of the edits first and sort that out once we're settled.

IJW-Wartrader: Looks good to me.

modules/main-rules/pages/transport.adoc
```diff
@@ -1,39 +1,61 @@
...
+== Transportation by units that have the {transport} ability
+
+Units that have the {transport} ability can carry other units across the battlefield.
+They can only transport Infantry units -- except where the rules or army lists say otherwise.
+
+To embark or disembark a unit:
+
+* You must sacrifice 5 cm from the movement of the {transport} unit.
+* You must sacrifice 5 cm from the movement of the unit that you embark or disembark.
+* You may embark and disembark units in the same turn, the same phase or even in the same move -- so long as all of the involved units have enough movement allowance.
```
IJW-Wartrader, line 19: What's a 'movement allowance'?

modules/main-rules/pages/transport.adoc
```diff
@@ -1,39 +1,61 @@
...
+====
+
 == Destruction of {transport} units
 
-If a {transport} unit is destroyed then, just before you remove the model as a casualty, roll a D6 for each embarked unit:
+If a {transport} unit is destroyed then, just before you remove the unit as a casualty, roll a D6 for each embarked unit:
 
-* *1–3*: The embarked unit is destroyed along with the {transport} unit.
+* *1–3*: The embarked unit suffers a hit.
 * *4–6*: The embarked unit survives -- place it in contact with the destroyed {transport} unit.
```
IJW-Wartrader, line 42: This is in reference to the two results of the die roll.

At some point (probably not now), this needs to be restructured so that placement of survivors is not linked to a specific result, but applies to anyone who survives. This will stop units with Save or multiple Damage Capacity from needing extra text.

That's partly why I would have liked to swap it from '4+ to live' to '4+ to get hit', as it allows for a much more concise structure like this:

... roll a D6 for each embarked unit. On a roll of 4+, the unit suffers a hit. If the unit survives, place it in contact with the destroyed {transport} unit.

modules/main-rules/pages/transport.adoc
digi-brain: Pretty much a cop-out until we've finished honing the whole 'move' thing. Potentially 'movement' is more than a single 'move', so your 'movement allowance' is the total of all the moves you're doing right now, if it is more than one move (ie, March orders).

Feel free to suggest an alternative that works.

modules/main-rules/pages/transport.adoc
```diff
@@ -1,39 +1,61 @@
...
-* You can simply embark and disembark units as you move the vehicle -- there is no cost to the {transport} unit's movement.
-But you may not make any independent movement with the Ork infantry units in the same phase in which they hitch a lift.
-* A unit of Orks can only hitch a lift once per phase.
+To use the {hitch} ability to transport an Infantry unit:
 
-+[TODO: Consider whether to move this 'Orks only' rule to the special rules statements in the Ork army list. 
-Unless any later content expanded it to other races?
-Also, might need to add detail to clarify 'multiple move' situations (March orders).]+
+* You must embark *and* disembark your Infantry unit in the same phase.
+Embark the unit at any point during your move of the {hitch} unit, and disembark either at the end that same move -- or at the end of another move in the same phase, if the {hitch} unit is on March orders.
```
IJW-Wartrader, line 54: This is a rules change, there are no restrictions on when the unit can disembark.

Maybe:

```suggestion
Embark the unit at any point during your move of the {hitch} unit, and disembark at another point during same move -- or during another move in the same phase, if the {hitch} unit is on March orders.
```

modules/main-rules/pages/transport.adoc
IJW-Wartrader: > Feel free to suggest an alternative that works.

'Movement'? 😜

More seriously, it's what the rest of the page is using for that concept, which is why 'movement allowance' stood out so much.

IJW-Wartrader: But it's a very minor issue.

modules/main-rules/pages/transport.adoc
digi-brain: >Don't forget that the unit is already sacrificing 5 cm of move just to be able to disembark.

I always rationalised the 5cm cost as a cost in **time** to embark/disembark, which seems reasonable and logical to me - not a 'compensation' for the width of the base, as you appear to see it. Though I can see that perhaps that makes sense from the point of view of game mechanics, but then why pay it twice?

But yeah, okay, let's skip this. The very fact that this part of the rules is controversial and creates arguments is exactly why I'm thinking that perhaps we could at least suggest 'ways to play this to make it more consistent or fair' -- but that could be in another section in future perhaps, especially as it might be hard to make concise, as you say.

It really irritates me when people 'conveniently' gain extra move for their units by virtue of the width of their base - so personally I tend to play it so that no individual unit can move farther in total than the maximum you can calculate that they should move. That seems the only fair way to me, and if that means moving your Battlefortress back a bit then so be it. If the Orks are massively disadvantaged in some way, then fix that rather than focus on loss of a gain that wasn't earned anyway.

But yes, different systems, different views. I've taken a view that seems the most consistent and fair to me.

I'll resolve this as 'nothing to do here', but if we do ever do a section for 'suggestions you can pick and choose for things that might otherwise be contentious or that don't render down to nice concise rules' then I think this is a contender for more attention!

modules/main-rules/pages/transport.adoc
digi-brain: Funny thing - I **swear** I saw something about 'at the end of the move' when I was revising the text, and that's why I worded it that particular way. But looking at the E40k source I don't see where it could've crept in from, if indeed it was there (I'll look at the history tomorrow just to satisfy my bafflement). I'm off to bed shortly, but good catch, I'll check your suggestion and fix tomorrow.

modules/main-rules/pages/transport.adoc
digi-brain: Yeah, I might've had some nuance in mind that set me on a tangent, but I can't think of it now. Maybe a brain fart. Movement will do. :-)
```suggestion
* You may embark and disembark units in the same turn, the same phase or even in the same move -- so long as all of the involved units have enough movement.
```

modules/main-rules/pages/transport.adoc
digi-brain: I see what you mean. Plus, in general I find it much easier to write such rules in a consistent way if we phrase such that the opponent rolls for this type of 'bad stuff' against you. Then the normal '4+/high numbers are good' expectation is met because it is 'good' from your opponent's point of view, just as it is when they roll to hit with their own shooting. Which you've probably thought of too. After all, it doesn't change the actual mechanic, it's still a 50:50 roll.

So, I agree it'd be nice not to hold up this PR for it, but I'm also supportive of the idea that we could try a rewrite of this bit later to 'invert' the result if it will make other elements more tidy and consistent.

modules/main-rules/pages/transport.adoc
digi-brain: Aha! It seems that I'm not entirely mad! :-D Previously, as from commit [e9e402b](https://github.com/GameBrains/er-core/pull/8/commits/e9e402becb9048a6b0c17e513b5f8ba133d158bb), it said:

`* Any embarked unit *must* disembark at the end of the {hitch} unit's move, or at the end of its last move during a March order.`

And... that was actually your commit, you bugger! ;-D

Prior to that it said:

`* You can simply embark and disembark units as you move the vehicle -- there is no cost to the {transport} unit's movement.`

Ah well, stuff happens. I'll look at your latest suggestion next.

IJW-Wartrader: > And... that was actually your commit, you bugger! ;-D

Doh! Oops. :-(

modules/main-rules/pages/transport.adoc
digi-brain: Of course, an option that would slot straight into the existing text:

```suggestion
On a roll of 1 to 3, the unit suffers a hit.
If the unit survives, place it in contact with the destroyed {transport} unit.
```

modules/main-rules/pages/transport.adoc
digi-brain: Done, so I'll resolve this. For future reference though, the pull request basically sees the renamed file as 'new' in comparison with the master branch it is comparing to. That's logical of course, but it makes it harder to see what is truly different from master, and it messes with the conversations and suggestions...

We do need to be free to rename files to match titles, but it seems that we also need to be aware of these limitations so we don't get into a pickle with any complex conversations or suggestion that might be in progress. As it is in the case there were only a couple of points to round off.

digi-brain (approved):
Ready to go - just sorting out merge conflicts.

PR #13, opened 2021-06-15
War engine barging

IJW-Wartrader:
I think this is ready to be looked at.

Assault moves and charges has had a major rewrite to use the updated rules logic. Note that:

- All the restrictions about 'moving past' _should_ now included in the main rules text.
- Charge moves are still listed, but are strictly a unit-level concept and are a subset of assault moves. We shouldn't need to update other pages that use 'charge' or 'charge move'.
- Example diagrams have been added. I'm not sure if the text for these is too long.


modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -2,36 +2,39 @@
...
-== Charge
+War engines follow the same rules for assault moves, but their own movement rules take priority. War engines can choose not to move in the Assault phase, but will not be able to initiate Close Combats or Firefights if they do not move.
+////
 
-As you move detachments in this phase, you may declare that any or all of them will *charge* so that they become engaged in close combat with the enemy.
+== Charge moves
 
-This is the only time that you may move your units into base contact with enemy units.
-And where you do move any individual unit into base contact with an enemy unit, we call this a *charge move*.
+If your unit's assault move brings it into base contact with an enemy unit, this is referred to as a *charge move*, or a *charge*.
```
digi-brain, line 33: I'm making some edits to this topic at the moment, mostly just different rephrasing. However, I think that this line is slightly problematic.

You're quite reasonably trying to construct things in such a way that we make assault moves and then state which of these are charges by definition because the units reached base contact.

However, I think that there is a bit of a 'chicken and egg' situation here. That is, I don't think that this construction works well given the special charge Speed bonuses for infantry, cavalry, etcetera. You only get their enhanced Speed value if you **will** make base contact.

So it isn't that "your unit's assault move brings it into base contact" per se (though that can be true), it is that the sub-set of assault moves that we call the charge move that brings it into base contact — which usually amounts to the same thing, but not always because of the special requirements for the bonus Speed.

I think that I'll finish my current phrasing edits and push them, and then we should look at bits like this — it get's really confusing if we're both making independent edits and suggestions at the same time, while also trying to compare our working copies with previous iterations.

digi-brain: Hang on, you go on to qualify it, so maybe it works out. Still, I might have to do the rephrasing in several iterations to wrap my head around it.

IJW-Wartrader: > Hang on, you go on to qualify it, so maybe it works out. Still, I might have to do the rephrasing in several iterations to wrap my head around it.

Yeah, I forgot to add in a note asking you to check these bits in particular, as even with the later text it's still slightly back-to-front in the order in which its explained...

digi-brain: I'll resolve this conversation — we know that it'll take a few passes to get through the whole topic and to thrash out the 'chicken and egg bit' in particular.

IJW-Wartrader:
@digi-brain out of curiosity, what was your reasoning behind intermingling the two barging options? Players will choose one option or the other before starting a game, so they'll only want the text of the option they're currently using.

digi-brain:
Mainly:

- It is easier for us to edit and maintain them in one list and avoid a mistake.
- Readers don't have to read both lists closely to check that we haven't made such a mistake.
- It's easier to compare.
- We don't need to use and explain the italics (though I did have to insert something to explain the two choices a bit instead).

IJW-Wartrader:
> * It is easier for us to edit and maintain them in one list and avoid a mistake.

I may not have explained the longer-term plan, sorry. There are _temporarily_ two options, one of which will be removed. I'm already 80% sure that the 5 cm version is the option that will be kept. Whether the 'final' option is then kept in the experimental/optional rules section or moved into the main rules (with the usual edition note) is something we can obviously decide later, after seeing what the community reaction is like. Although it's yet another rules change, it allows behaviour that the rules examples clearly expected to be allowed.

There isn't going to be a prolonged period of editing and maintaining two options.

> * It's easier to compare — and it's easy to gloss over the bit you don't need.

I disagree, strongly. Bear in mind that you (aka the players) aren't going to be constantly comparing the two options. You'll read through the two options and pick one version to test, and from that point on you don't need to look at any part of the other option.

To be honest, I was already wondering about removing the italics and note completely. I wasn't even sure it was a good idea to put both options on the same page!

modules/experimental/pages/war-engine-barging.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+.Option A: Barging based on initial Damage Capacity (DC)
+[start=2]
+. Rather than stop there, you may continue your move if you have more movement available.
+. You may not continue your move past the point at which you run out of available movement or the War Engine has made contact with a number of enemy units equal to half its initial Damage Capacity -- whichever happens first.
+
+.Option B: Barging based on distance (5 cm)
+[start=2]
+. Rather than stop there, you may move your War Engine up to a further 5 cm, or the remainder of its available move -- whichever is the least.
+ * Remember that a War Engine unit can attack or be attacked by no more units than the number equal to half its initial Damage Capacity. So, you must end your assault move if you reach this condition before you run out of movement.
```
IJW-Wartrader, line 37:
```suggestion
So, you must end your assault move if you reach this condition before you run out of movement.
```

This isn't a pull request suggestion, but picking out a specific bit of text to comment on.

The original text for the 5 cm option doesn't limit the number of units that can be barged out of the way.

Is this an intentional change to the option? I wondered about adding this as a third restriction, but decided to leave it out to reduce complexity. The note was included as a reminder that even if more than DC/2 units end up in base contact, they won't all be fighting in close combat.

digi-brain:
>There are _temporarily_ two options, one of which will be removed.

Yes, I understood that. We just make some quick edits again once we know which option we'll keep (or split).

>There isn't going to be a prolonged period of editing and maintaining two options.

Maybe not, but you never know, and it is still good to practice DRY (don't repeat yourself) coding in the meantime. And even editing it **once** was fiddly because there are bits in common across two similar sections that you have to digest and keep in common...

Also, people might want to compare more than you think while they decide which they prefer — we're giving people two options, **of course** they will compare them, and maybe not only the one time.

> > * It's easier to compare — and it's easy to gloss over the bit you don't need.
> 
> I disagree, strongly.

You disagree that it is easier to compare, and disagree that is easy to gloss over the bit you don't need, strongly? I honestly can't see the problem or why it should stir a strong response.

>I wasn't even sure it was a good idea to put both options on the same page!

When there is so much commonality I think it is an **eminently** good idea to put them both on the same page **at least** while we're thrashing it out, for everyone's sake -- authors, editors and readers.

If you think that there is some strong reason why one list is a terrible idea then I think we'll have to discuss over a video chat, because I'm not seeing the problem, at all. :)

modules/experimental/pages/war-engine-barging.adoc
digi-brain: >The original text for the 5 cm option doesn't limit the number of units that can be barged out of the way.

With respect, yes it did. The 5 cm option section ended with a note that included this text:

>_Remember war engine units can only attack or be attacked by units equal to half the war engine unit’s starting Damage Capacity._

I've just moved it up, integrated it with the procedure and reworded it a bit. The previous note that came after both procedures also took some parsing because it wasn't clear whether it related to only the last procedure, or both (and was in fact a mix because it also had "Enemy war engines *cannot* be barged aside.").

This was another reason why I created one procedure with two optional clauses rather than leave it as two very similar procedures with 'other stuff' near them where the relationship is less clear.

IJW-Wartrader:
> If you think that there is some strong reason why one list is a terrible idea then I think we'll have to discuss over a video chat, because I'm not seeing the problem, at all. :)

Yes, I think we'll have to. As a player, once I've chosen one version of a playtest rule for a game, I want to read _that_ version of the rule during the game, not faff around mid-game disentangling two intermingled texts to find the one I want. This is not hypothetical, it comes from direct experience.

For the playtest stuff, I need more control than normal for what gets published. _After_ playtesting is another matter.

modules/experimental/pages/war-engine-barging.adoc
IJW-Wartrader: > > The original text for the 5 cm option doesn't limit the number of units that can be barged out of the way.
> 
> With respect, yes it did. The 5 cm option section ended with a note that included this text:
> 
> > _Remember war engine units can only attack or be attacked by units equal to half the war engine unit’s starting Damage Capacity._

I might be missing a rule somewhere, in that case. I'm not aware of a direct link in E40k between how many units are allowed in base contact, and how many are allowed to attack/be attacked. 

It's Epic Armageddon that limits the number of units that can enter base contact.

digi-brain: Possibly at cross purposes again, but...

While I only rewrote what you already wrote, it did sound familiar and I double checked it to be sure:

>Ordinary units can only be ganged up on by a maximum of two enemy units, but war engines can be attacked by a number of units equal to half of their starting Damage Capacity. Units attacking a war engine in close combat must be in contact with the war engine model or its base.

https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/WarEngines/WarEnginesInCloseCombat.htm

So it's not just EA, unless there is some nuance in there that you're referring to?

digi-brain: Ah, you're talking about number in base contact vs number that are actually attacking. Hmmm... that seems a bit obtuse and needlessly messy.

IJW-Wartrader: > Ah, you're talking about number in base contact vs number that are actually attacking. Hmmm... that seems a bit obtuse and needlessly messy.

That, and 'Add up Assault values':

> Each player starts by adding up the Assault values of units in base-to-base contact with the enemy. A maximum of two units can 'gang up' on each enemy unit.

I think there's a _very_ strong case for rewriting the main rules to limit it by base contact and moving that restriction to the charge moves section so that it's not being checked _after_ all the charge moves have been made, but we haven't done that yet and I wasn't sure if that was why you'd 'changed' the text.

Note that I don't mind the change, but I wasn't sure if it was because you thought the game logic be changed or because your brain had sensibly over-ridden the obtuseness of the GW text.

IJW-Wartrader:
Just to confirm, the assault moves and charges content has been passed over to you for technical writing.

The war engine barging page is still playtest content and is not in a finished enough state to hand over.

I'm not sure of the best way to handle mixed content that I've ended up putting in the same branch. In this case, maybe it would be easiest if I redo the barging page, keeping the best of your rephrasing, and then it can be committed once you're happy with the assault and charge move parts.

IJW-Wartrader:
I've made changes to the barging page, in two commits to make it easier to see different changes.

I've kept the rules 'change' to restricting the number in base contact rather than mentioning how many can fight, which has allowed the whole thing to be restructured as _one_ procedure, and Option B is now just an additional restriction to apply.

The second commit then changes the 'stop your barge if X happens' part to a bullet point for each restriction, turning the note about enemy WEs into a restriction bullet. This also makes Option B easier to write and understand, as it becomes an extra restriction bullet.

Could you please have a look at Option B, and see if you can think of a better way to describe '5 cm of continued movement'?

digi-brain:
>Could you please have a look at Option B, and see if you can think of a better way to describe '5 cm of continued movement'?

Frankly, it does not work well when you break it into two sections like this. You're forcing the reader to find another option after the procedure, navigate back up to the procedure, integrate a bullet point into that procedure, and parse the lead-in sentence with the new bullet point alongside the others.

I'm not too fazed in this case, given that it is experimental/playtesting content, but...

...if you're going to ask me to justify every other edit I make "out of curiosity", then reject all of my reasoning and decide that what I did was terrible and that you know better, and then replace it with something objectively worse like this... then we have trouble ahead. Just sayin'.

Let's discuss soon. I'm happy to have a looser approach for experimental/playtesting stuff, I can see the sense of that. But I'm concerned about how we're going to cope with other edits.

I'm a bit busy this week, but I'll ping you sometime in the next couple of days on the off-chance that you're free, or think about a time slot soon.

modules/experimental/pages/war-engine-barging.adoc
digi-brain: >Note that I don't mind the change, but I wasn't sure if it was because you thought the game logic (edit:) should be changed or because your brain had sensibly over-ridden the obtuseness of the GW text.

Okay, I don't mean to labour the point, but you still seem to be under the misapprehension that I changed your meaning: That can happen, but as mentioned above, in this case I do not believe that it did — if there is a 'change', it stems from your preceding version in which you wrote:

>Remember war engine units can only attack or be attacked by units equal to half the war engine unit’s starting Damage Capacity.

Having cleared that up, I think that you are right that, strictly speaking, GW's text is ambiguous and could make people think that the number of units in base contact and the number that can actually attack/be attacked are two separate things. But, unless there is other evidence to corroborate it, I think that that is a pretty obtuse interpretation that rests on GW's poor wording rather than actual design intent.

Given how messy that interpretation appears to be, I agree that we should 'override' it (or rather, in my opinion, clarify so that there is no longer room for the erroneous interpretation).

digi-brain: Per video chat, I did actually get double-blind on this one, so sorry about that.

We seem to agree that the other interpretation is a bit obtuse and messy, and that we should take the simpler interpretation in which there is a 1:1 relationship between base contact and number of attackers/targets. So, I'll resolve this conversation.

digi-brain (approved):
@IJW-Wartrader I've completed my current editing pass on this. `assault-moves-and-charges.adoc` is the trickiest bit by far, but I think it hangs together okay. So, I feel happy to go ahead and merge this into master branch to publish — assuming you don't spot any rules errors I might've introduced in the rephrasing and restructuring of course, as always.

Per our video chat about process for experimental/playtest stuff, we disagree about presentation of the two variations within `war-engine-barging.adoc`. However, while it is in the experimental section I'm happy to agree to disagree and let you structure it the way you prefer. You can always adjust it more in future of course, but for the moment do you want to adjust it some more per your preference or do you want to publish it as it stands now?

IJW-Wartrader:
> Per our video chat about process for experimental/playtest stuff, we disagree about presentation of the two variations within `war-engine-barging.adoc`. However, while it is in the experimental section I'm happy to agree to disagree and let you structure it the way you prefer. You can always adjust it more in future of course, but for the moment do you want to adjust it some more per your preference or do you want to publish it as it stands now?

I'd like to adjust it once more (this evening), and then publish.

IJW-Wartrader:
> > Per our video chat about process for experimental/playtest stuff, we disagree about presentation of the two variations within `war-engine-barging.adoc`. However, while it is in the experimental section I'm happy to agree to disagree and let you structure it the way you prefer. You can always adjust it more in future of course, but for the moment do you want to adjust it some more per your preference or do you want to publish it as it stands now?
> 
> I'd like to adjust it once more (this evening), and then publish.

Adjustments done!

digi-brain:
Cool.

For future reference, I think that the IMPORTANT note is a bit redundant — much the same intent is implicit in the very next sentence "An experimental option..."
Also, I intend to add something general to the top level experimental section topic, which will make it even more redundant.

And, be aware that it is best not to start a topic with anything that might hog the limelight in search results ahead of the introductory summary (short description).

However, I'll just leave it alone right now and publish as-is. Looks good as rules to playtest.

Issue #11, opened 2021-06-09
Transportation - how it might really work, and base size issues


IJW-Wartrader:
I've pulled this out into an issue so that we can carry on discussing it separately from the special abilities branch, and included all three relevant comments...

---

Eric:

As written, might this permit cases where players could have sacrificed some movement, but don't because the rules say they can put the units in base contact anyway - so they use that to get a 'free' gain?

In general, I feel like perhaps there should be some concept of not exceeding the 'calculated maximum move you could get from the transport plus the embarked unit in combination' -- even if this means you must move transport unit a bit less or something to compensate rather than adding on the width of the disembarked unit's base and thus going beyond the calculated maximum move.

If there is something in this, I'd be happy for us to create an issue to tackle this properly later - just wanted to raise the point and get your thoughts.

_Originally posted by @digi-brain in https://github.com/GameBrains/er-core/pull/8#r648571469_

IJW-Wartrader:
> As written, might this permit cases where players could have sacrificed some movement, but don't because the rules say they can put the units in base contact anyway - so they use that to get a 'free' gain?

As far as I can see, this would require them to have a base longer than 5 cm. 

Don't forget that the unit is **already** sacrificing 5 cm of move just to be able to disembark. This is another reason why I automatically assumed that you 'deploy then measure' - as far as I was concerned, the unit had paid a movement penalty to get out, why would the unit have to pay a **second** movement penalty by measuring from the transport to 'move' out?

As a reminder, don't forget that what we currently have is already going to be regarded as a controversial rules change by some players, and if they've played 'deploy then measure' for a long time it's going to be unpopular, at best.

> In general, I feel like perhaps there should be some concept of not exceeding the 'calculated maximum move you could get from the transport plus the embarked unit in combination' -- even if this means you must move transport unit a bit less or something to compensate rather than adding on the width of the disembarked unit's base and thus going beyond the calculated maximum move.

My gut reaction is that this part of the map is marked 'Here be Dragons'. It's likely to be hard to write clear concise rules for, the overall subject is already controversial/generates lots of arguments, it's unlikely to get widespread support from the players (and that includes me), and it adds even more inconsistencies with Hitch. Or you apply it to Hitch as well, and screw over the Orks even more.

Paraphrasing...

'You know that unstoppable Battlefortress war engine that isn't slowed down at all when Da Boyz leap suicidally off the side?'
'Yer.'
[...]

PR #18, opened 2021-08-07
Transportation

IJW-Wartrader:
Transportion page updated to include embarking and disembarking completely within 5 cm, and so that destruction of transportation units applies to both types of transportation - because a Hitch unit could die during its movement to a Dangerous terrain test, minefield etc.


modules/main-rules/pages/transportation.adoc
```diff
@@ -2,12 +2,42 @@
...
+====
+We added this rule to minimise the effects that result from the assortment of different base sizes that players use.
+====
+
+== Destruction of transportation units
+
+If a transportation unit is destroyed then, just before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the unit suffers a hit.
+If the unit survives, place it completely within 5 cm of the destroyed transportation unit.
+
+Then remove the transportation unit as normal.
```
digi-brain, line 28: Am I right to think that this situation doesn't actually arise with Hitch (X) units, except in the rare case of a failed Dangerous Terrain test followed by another roll of 1...?

If so then I think we should clarify this. And, this would strengthen the argument for us to move the Transport(X)-specific and Hitch(X)-specific sections over to the respective topics, given how little 'common' content that would leave us with for this topic.

In other words, it might be better to restore the 'Destruction of {transport-x} units' section or similar, and match it with a 'Destruction of {hitch-x} units' section that clearly shows this difference.

IJW-Wartrader: Yes, Dangerous terrain tests, but if you're using lots of Hitch units and terrain it comes up every other game or so. However what's more common if you're using games with hidden setup is booby traps (hit on a 2+) and minefields (roll to hit every moving unit).

But that makes no difference to how the 'destruction of x' rule works, it's identical for both types of transportation unit. Maybe just a 'tip' note pointing out when Hitch units can be destroyed while carrying someone? Or an edition note saying that E40k didn't say what to do when a Hitch unit was destroyed while carrying troops.

digi-brain: Okay, that's clear. The reason that I suggest a possible split here is that something like these two similar but different passages:

> If a Transportation (X) is destroyed then, just before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the unit suffers a hit.

> If a Hitch (X) unit is destroyed while it is motion then, just before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the unit suffers a hit.

...may turn out to be neater than a single passage in which we try to spell out the differences, with or without a note. Especially if we do find it better to move the ability-specific stuff to the ability topics and leave only the most generic stuff in the transportation topic.

I'm not sure though. I'll have a play with it soon, see how it looks.

IJW-Wartrader: I think I'm missing something. What's the purpose of having two different passages? If a transportation unit is destroyed with units embarked on it, the game mechanics are identical.

'while it is in motion' definitely can't be used instead of a reminder of how the units can be destroyed, because everything that can kill a moving Hitch unit can also kill a moving Transport unit.

digi-brain: >'while it is in motion' definitely can't be used instead of a reminder of how the units can be destroyed, because everything that can kill a moving Hitch unit can also kill a moving Transport unit.

I think you're sort of looking at it from the opposite end to me.

As I understand it, units are only 'embarked' on a Hitch (X) unit while it is in motion. We embark them at some point during the Hitch (X) unit's move, and disembark them at some other point in its move (or sequence of moves) -- often at the end. If the Hitch (X) unit is destroyed when it is at rest then there are no embarked units that might get hit -- which is different to a Transport (X) unit which may contain embarked units while it is at rest.

Hence why the 'in motion' bit might work as an elegant way to differentiate, while the alternative might be more clunky. The rest of the passage to determine hits works the same, sure, but the conditions to trigger this differ because they are more limited for Hitch (X).

IJW-Wartrader: > Hence why the 'in motion' bit might work as an elegant way to differentiate, while the alternative might be more clunky. The rest of the passage to determine hits works the same, sure, but the conditions to trigger this differ because they are more limited for Hitch (X).

My issue is that including the 'in motion' bit for Hitch units infers that Transport units that die 'in motion' _don't_ follow these rules. Dying while in motion is a condition that applies to both types of transportation unit, so it can't be used as a differentiator, only as a reminder.

I don't think it's necessary to spell out the ways in which a transportation unit can die, but the text could be tweaked to reinforce it only applying while carrying other units:

```suggestion
== Destruction of transportation units

If a transportation unit is carrying other units and is destroyed then, just before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the unit suffers a hit.
If the unit survives, place it completely within 5 cm of the destroyed transportation unit.

Then remove the transportation unit as normal.
```

If you want to spell it out,  a reminder note could be added to the {hitch-x} section of the page, for example:

>TIP: As {hitch-x} units can only carry units while they are moving, the _Destruction of transportation units_ rule only applies if the {hitch-x} unit dies during a move, for example to a failed Dangerous terrain test, or being hit by a Booby-trap or Mine-field.

digi-brain: >My issue is that including the 'in motion' bit for Hitch units infers that Transport units that die 'in motion' don't follow these rules

Why should it infer any such thing? The 'in motion' bit is only for the Hitch (X) passage. The equivalent Transport (X) passage doesn't have the 'in motion' condition, so by default it applies to all conditions. That is literally the point.

There is more than one way to skin a cat, and I haven't experimented with it yet to see how it actually looks — but the point here is that those few words around 'in motion' might be more elegant than a big 'note' paragraph like the one you suggest. Unless you think that it is too subtle and would still need further explanation. But that's a different issue to inferring the wrong thing, which I can't see that it does, as above.

IJW-Wartrader: > Why should it infer any such thing? The 'in motion' bit is only for the Hitch (X) passage. The equivalent Transport (X) passage doesn't have the 'in motion' condition, so by default it applies to all conditions. That is literally the point.

Hmm, let's start again.

Being destroyed is a trigger. Being in motion is not part of that trigger, but for Hitch units it's a prerequisite to _reach_ the trigger of being destroyed.

The text conflates the trigger and the prerequisite for that trigger, which means that the reader can't take the text at face value. They have to analyse the text to work out why the concepts are being conflated. And that can result in reading the passages like this:

If X is true, apply Z to Transport units.
In the separate situation where X _and_ Y are true, apply Z to Hitch units instead.

I think we can agree that we don't want the reader to even _suspect_ that this is the game logic!

----

Bear in mind that I think we're going to need an edition change note anyway, as destruction of Hitch units while carrying other units is missing in the original rules. If you want to spell out that Hitch units can only carry units while moving, so embarked units can only die during the move, and we need a note anyway, why not mention it in the note?

----

As an aside, and I know this is encroaching on your territory, is it a good idea to have all the transportation stuff split across three pages? I'm a bit concerned that the original rules forced the reader to flick back and forth in the book to work out how Transport (x) worked, and we're currently doing the same.

Might it be worth pulling the small amount of info from hitch-x.adoc and transport-x.adoc into transportation.adoc and leaving them as stubs, or expanding hitch-x.adoc and transport-x.adoc into complete entries with a partial for the parts that are the same for both, with transportation.adoc becomign a stub?

digi-brain: I'm still not seeing whatever it is that you're seeing. But it was only an idea I wanted to check with you— I'll drop this specific iteration of it, as it seems to be failing that check.

---

>Bear in mind that I think we're going to need an edition change note anyway

Yes, but that means **two** notes, one of which we need no matter which way we do it — one for the actual clarification/change itself, and one to highlight the fact that it is a change from E40k.

>why not mention it in the note?

I don't believe it is a good idea to conflate the rule or clarification — that we would normally put in the main text or in a regular note — with the purpose of the 'E40k difference' note. The latter highlights and explains the former; it doesn't take its place.

---
>is it a good idea to have all the transportation stuff split across three pages?

It depends on how it is split, as you've alluded to yourself. There's more than one possible approach, and I haven't got a clear idea of which is likely to work best from just my mind's eye.

>the original rules forced the reader to flick back and forth in the book to work out how Transport (x) worked, and we're currently doing the same.

I'd say we're not, currently. The Hitch (X) and Transport (X) ability topics barely say anything at all — the real content is in the Transportation topic. However, I am wondering if it might work better to move the content to the ability topics, when there is so little that is truly in common anyway. I recall that you also pondered this very idea previously — I just figured that now might be the time to give it full consideration, as an alternative to deleting the line that you commented to the effect that it 'isn't really true' because it says that the detail is in the ability topics when in fact it isn't at present.

As I said, I need to experiment with it a bit I think. But I suspect that `stub Transportation topic + two short detailed ability topics` will work better than `1 long detailed Transportation topic + two stub ability topics`. For me, your confirmation that the conditions are indeed different reinforces this suspicion, because it leaves even less that is truly in common and without caveats between the two modes of transportation.

modules/main-rules/pages/transportation.adoc
```diff
@@ -2,12 +2,42 @@
...
 We explain these and other special abilities in more detail elsewhere; but transportation is a vital aspect of the Movement phase so we'll discuss it here too.
+// IJW editing note - the second sentence isn't really true, this page is where  the detail is!
 
 When we speak of these transportation abilities, we use the terms 'embark/embarked' and 'disembark/disembarked' to describe actions and states that relate to the units that you transport.
 
 Both the {transport-x} and {hitch-x} abilities indicate units that can carry -- transport -- other units across the battlefield.
 But they work in slightly different ways.
 
+== Placement of units
+To embark a unit, it must be completely within 5 cm of the transportation unit.
```
digi-brain, line 13: Is this section meant to cover both Transport (X) and Hitch (X), as it appears from the context?

If so then it seems to imply that Hitch (X) units no longer have to make contact with the units they will carry — only get close enough so that those units are fully within 5 cm. Is this right? I mean, movement is sacrificed for Transport (X) so it balances out, but that's not the case for Hitch (X) — so doesn't this change imply a significant boost to Hitch (X)? Or am I missing something that balances it out?

If not, what **does** apply to Hitch (X) as far as evening out base size issues for embarking and disembarking are concerned?

IJW-Wartrader: > Is this section meant to cover both Transport (X) and Hitch (X), as it appears from the context?

Yes. As per https://github.com/GameBrains/er-core/issues/11#issuecomment-869613462 the main purpose of this branch is to apply the same rules for embarking and disembarking (apart from the different costs) to Hitch and Transport, in a way that reduces the impact of differing base sizes.

> If so then it seems to imply that Hitch (X) units no longer have to make contact with the units they will carry — only get close enough so that those units are fully within 5 cm. Is this right? I mean, movement is sacrificed for Transport (X) so it balances out, but that's not the case for Hitch (X) — so doesn't this change imply a significant boost to Hitch (X)? Or am I missing something that balances it out?

Well...

_Some_ movement is sacrificed for Transport.
**All** movement is sacrificed for Hitch.

But that's not the issue - Hitch is almost always used to drop units off, and then pick them up from the same place during the next move of the Hitch unit. This means that embarking and disembarking need to be exact antonyms in game mechanics. So if units can be dropped off completely within 5 cm (to avoid base size issues), they need to be picked up from completely within 5 cm so that Hitch actually works.

In practical terms, when using the standard strip bases, you 'gain' exactly the same 1 cm per embark and disembark for Hitch as you gain for Transport, because the far end of the base is 5 cm away from the transportation unit, instead of 4 cm for placing them in base contact. Smaller bases 'gain' more, but again that's the purpose of all the 'completely within 5 cm' stuff.

modules/main-rules/pages/transportation.adoc
IJW-Wartrader: > > Bear in mind that I think we're going to need an edition change note anyway
> 
> Yes, but that means **two** notes, one of which we need no matter which way we do it — one for the actual clarification/change itself, and one to highlight the fact that it is a change from E40k.
> 
> > why not mention it in the note?
> 
> I don't believe it is a good idea to conflate the rule or clarification — that we would normally put in the main text or in a regular note — with the purpose of the 'E40k difference' note. The latter highlights and explains the former; it doesn't take its place.

Can we start this discussion from scratch, from first principles?

- A Hitch unit moves, embarking and then **always disembarking** carried units during the same move/march.
- If the Hitch unit gets shot, or hit in close combat etc, it's impossible for it to be carrying other units at that time.
- Players never need to look up what happens if a stationary Hitch unit is destroyed while carrying units, because that situation cannot occur during a game.

If the game state can never include a situation, we don't need additional rules text to exclude that situation. 

As far as I can see, this means that all we need is an edition change note, saying that we've expanded the rules to cover Hitch units as they can die while carrying units.

digi-brain: >* Players never need to look up what happens if a stationary Hitch unit is destroyed while carrying units, because that situation cannot occur during a game.

Um, yes... good point. While I recognised the difference, I think I had some sort of idea that players who are used to Transport (X) might fall foul of 'muscle memory' and apply the same conditions to Hitch (X) -- forgetting that most of the time hitching units aren't actually embarked (only actually embarked while 'in motion'). I still think that could happen quite easily; but now I take your point that it warrants only a clarifying note at most.

Okay, good -- on the same page at last, I was just a bit slow/side-tracked on this one! :-D

modules/main-rules/pages/transportation.adoc
digi-brain: >*Some* movement is sacrificed for Transport. **All** movement is sacrificed for Hitch.

It doesn't seem to be quite that straightforward to me. To be precise:

**In E40k**:
* For Transport (X), both the troops and the transport units sacrifice some movement.
* For Hitch (X), only the troops sacrifice movement — the transport units only sacrifice movement in as much as they have to deviate from their optimum course in order to embark the troops.

**But, in this draft for ER**: 
* It seems to me that we're effectively saying that Hitch (X) units now make even less sacrifice from the optimum course, and troop units *potentially* gain a bit. That is, they won't gain in the standard 'deployed and moving across the board' case, but they will gain in some other situations that are more typical of **re**deployment.
* This is because the transport only has to get close enough to case the troop unit to be fully within 5 cm — which in many cases will mean they only have to get within 3 to 4 cm.

So, have I identified an **actual** problem here that demands a bit more thought, or is it balanced out, insignificant or irrelevant because of some other part of the equation that I haven't considered?

IJW-Wartrader: > * It seems to me that we're effectively saying that Hitch (X) units now make even less sacrifice from the optimum course, and troop units _potentially_ gain a bit. That is, they won't gain in the standard 'deployed and moving across the board' case, but they will gain in some other situations that are more typical of **re**deployment.

Yes, in exactly the same way that Transport (X) units make slightly less sacrifice from their optimum course in those situations. Remember that the primary aim for this branch is rationalising Transport (X) and Hitch (X), with rules changes from E40k because the polls about this were overwhelmingly in favour of updating the transportation rules.

> * This is because the transport only has to get close enough to case the troop unit to be fully within 5 cm — which in many cases will mean they only have to get within 3 to 4 cm.

No, it won't. 4 cm would require the transported unit's base to be 1 cm diameter*, and 3 cm would require the base to be 2 cm diameter. You can't even get an old square base completely within 5 cm and have it be 3 cm from the transportation unit.

*It could be oval and bigger on the longer axis, but oval infantry bases are incredibly rare.

> So, have I identified an **actual** problem here that demands a bit more thought, or is it balanced out, insignificant or irrelevant because of some other part of the equation that I haven't considered?

Embarkation and disembarkation have to be mirror-image manoeuvres to prevent weird results where you can drop off a unit but then can't pick up the dropped off unit in a subsequent phase, even though nobody has moved since then.

This makes _both_ forms of transportation slightly more flexible, but doesn't give either of them an advantage over the other.

If you have time at some point, could you give these rules a try on the game table? The need for embarkation and disembarkation to be mirrored manoeuvres gets a lot clearer in action.

digi-brain: >No, it won't. 4 cm would require the transported unit's base to be 1 cm diameter*, and 3 cm would require the base to be 2 cm diameter. You can't even get an old square base completely within 5 cm and have it be 3 cm from the transportation unit.

Okay, mostly true — outside of some arrangements with larger transports that have long rectangular bases — so we're talking something like ~2.5cm maximum in most cases when you account for the arc of measurement. (Now that I think of it, I tend to picture a simple arrangement with a perpendicular measurement in my head — but I should really picture an arc.)

But as long as both types of transport are getting the same overall gain, and the gain is slight enough in the overall scheme of things, then I guess it won't matter — which seems to be the crux of your points.

>If you have time at some point, could you give these rules a try on the game table?

Unfortunately, I'm unlikely to make time for that any time soon, unless my main Epic gaming friend pops up and we get an actual proper game in. In any case, I don't disagree that embarkation and disembarkation should mirror each other. I'm just querying other **potential** implications or disparities that occur to me as I edit the words.

In this case it seems that I **haven't** identified an **actual** issue that needs more thought — so, I'm happy to move on.

modules/main-rules/pages/transportation.adoc
digi-brain: I've pushed a bunch of edits, but there is more to do. Eg:

1. Some annotations to resolve.
2. It may yet be best to integrate the embark/disembark placement conditions into each set of transportation rules, rather than have them under a separate heading — I find it a bit disjointed.

Also note that, aside from rationalisation of text that defines what units each ability can carry, I have not split the ability-specific stuff off to the separate ability topics. I have misgivings about both approaches; but so far my experiments lead me to think that the current approach is more user-friendly, on balance. This does deviate from the approach we've taken with the likes of Artillery and Flak — but it is a bit of a special case.

I welcome your thoughts on this. Though of course we can always change it later if we have a brainwave, so no point in tying ourselves up in knots if the current approach is 'good enough'.

digi-brain:
I've pushed a bunch of edits, but there is more to do. Eg:

1. Some annotations to resolve.
2. It may yet be best to integrate the embark/disembark placement conditions into each set of transportation rules, rather than have them under a separate heading — I find it a bit disjointed.

Also note that, aside from rationalisation of text that defines what units each ability can carry, I have not split the ability-specific stuff off to the separate ability topics. I have misgivings about both approaches; but so far my experiments lead me to think that the current approach is more user-friendly, on balance. This does deviate from the approach we've taken with the likes of Artillery and Flak — but it is a bit of a special case.

I welcome your thoughts on this. Though of course we can always change it later if we have a brainwave, so no point in tying ourselves up in knots if the current approach is 'good enough'.

modules/main-rules/pages/transportation.adoc
digi-brain: Resolving this conversation as we reached a conclusion and it is now outdated.

modules/main-rules/pages/transportation.adoc
```diff
@@ -1,59 +1,80 @@
...
+To embark or disembark one or more units:
 
 * You must sacrifice 5 cm from the movement of the {transport-x} unit.
 * You must sacrifice 5 cm from the movement of the unit that you embark or disembark.
 * You may embark and disembark units in the same turn, the same phase or even in the same move -- so long as all of the involved units have enough movement.
 
-You may perform no other actions with units that are embarked -- so, you cannot shoot with them, for example.
+You may perform no other actions with embarked units -- so, you cannot shoot with them, for example.
 Obviously, once you disembark them then the normal rules apply again.
+// TODO: I think we might need to refine this because 'action' might be ambiguous -- for example, shouldn't the 'stubborn' part of the Supreme Commander ability still be allowed even while the unit is inside transportation, while the 're-roll Assault resolution die' part would not?
```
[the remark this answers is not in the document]
digi-brain: I don't see a problem there, I can't imagine that anyone would see it any other way. That's not what I mean to highlight.

My point is more fundamental: "You may perform no other actions" might be ambiguous for some readers.

Arguably, in a broad sense some abilities imply 'action' of some kind. So, might some readers take that statement to mean that special abilities are wholly inoperable while a unit is embarked within a transport unit?

Whereas in fact we intend that appropriate parts of abilities remain operable where there is not an obvious contradiction. The Supreme Commander example I gave highlights an instance where part of the ability would apply and part would not (because the SC isn't an active participant in the assault if he's inside the transport unit and not outside fighting).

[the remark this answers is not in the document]
IJW-Wartrader: > My point is more fundamental: "You may perform no other actions" might be ambiguous for some readers.
> 
> Arguably, in a broad sense some abilities imply 'action' of some kind. So, might some readers take that statement to mean that special abilities are wholly inoperable while a unit is embarked within a transport unit?

Have you got any examples? The only special abilities I can think of at the moment that involve an 'action' rather than a static ability is playing Psychic Blast Fate Cards from a Psyker, or Save stopping a hit.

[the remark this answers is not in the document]
digi-brain: Well, the one that sprung to mind was the re-roll part of Supreme Commander, which is why I gave it as an example. I didn't dig into it further, and I may have a misplaced nag from EA, but yes, part of the Psyker ability and perhaps the Medic ability might be construed as 'actions' — along with possible future abilities we haven't dreamed up yet.

I wonder if some readers — particularly if English is not their first language and they conflate concepts like 'action' and 'effect' — *might* construe that statement to cancel parts of an ability that should remain operable.

My note was more query than statement — I'm unsure and could be overthinking it. If, with the above in mind, you don't agree that the statement could be ambiguous, then I'm happy to drop it.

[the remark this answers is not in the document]
IJW-Wartrader: I think (hope) that this could be overthinking it!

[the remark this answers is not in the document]
digi-brain: Righto, I'll delete that note. It's not like we can't change it if it ever does prove to be problematic for some people.
```suggestion
```

modules/main-rules/pages/transportation.adoc
```diff
@@ -1,59 +1,79 @@
...
 On some unit data sheets you'll see that the unit has one of two special abilities that relate to transportation: {transport-x} or {hitch-x}.
-We explain these and other special abilities in more detail elsewhere; but transportation is a vital aspect of the Movement phase so we'll discuss it here too.
 
-When we speak of these transportation abilities, we use the terms 'embark/embarked' and 'disembark/disembarked' to describe actions and states that relate to the units that you transport.
+These abilities indicate units that can carry -- transport -- other units across the battlefield.
+These abilities have some features in common, but also differ in some important ways.
 
-Both the {transport-x} and {hitch-x} abilities indicate units that can carry -- transport -- other units across the battlefield.
-But they work in slightly different ways.
+== Embark and disembark
```
digi-brain, line 8: Just adding a note here to repeat something I mentioned in the main thread, so I don't lose track of it:

It may yet be best to integrate the embark/disembark placement conditions into each set of transportation rules, rather than have them under a separate heading — I find it a bit disjointed.

digi-brain: I'll resolve this conversation, as it is now outdated by my latest commit to address this.

digi-brain:
In my latest commit I've rationalised this some more; in particular, I've integrated the embark/disembark stuff that makes each section more self-contained and easy to follow. I've also introduced the term 'cargo' unit to make it easier to 'target' units in the rules more clearly where 'embarked' or 'disembarked' won't work.

I think its an improvement, but do please make a local build and check the flow and detail. I may add a couple of conversation comment in GitGub shortly too.

modules/main-rules/pages/transportation.adoc
```diff
@@ -1,59 +1,62 @@
...
+====
 
-Units that have the {hitch-x} ability can carry Infantry units across the battlefield -- they 'hitch a lift'.
-That is, they simply grab onto the outside of the vehicle and hang on for a crazy ride over the battlefield.
+== Transportation of units from another detachment
 
-To use the {hitch-x} ability to transport an Infantry unit:
+Where transportation units from one detachment carry units from another detachment, both detachments must be on the same orders.
+// IJW editing note - the war engine orders page will need a note saying how this works if you're *not* using war engine orders.
+// TODO: I wonder if we should thrash this detail out in the Experimental section? I mean, the original war engine orders rules were an afterthought and by no means universally accepted. Thus, I suggest that until such time as we nail down an alternative we should default to the original 'no orders' -- in which case this tract won't make sense here anyway.
```
digi-brain, line 53: Adding a comment here to draw attention to these notes, for further discussion.

modules/main-rules/pages/transportation.adoc
```diff
@@ -1,59 +1,62 @@
...
-We explain these and other special abilities in more detail elsewhere; but transportation is a vital aspect of the Movement phase so we'll discuss it here too.
 
-When we speak of these transportation abilities, we use the terms 'embark/embarked' and 'disembark/disembarked' to describe actions and states that relate to the units that you transport.
+These abilities indicate units that can carry -- transport -- other units across the battlefield.
+These abilities have some features in common, but also differ in some important ways.
 
-Both the {transport-x} and {hitch-x} abilities indicate units that can carry -- transport -- other units across the battlefield.
-But they work in slightly different ways.
+To describe the actions that relate to the units that you transport we use the terms *embark* and *disembark*.
+To describe the units that we embark and disembark, we use the term *cargo unit*.
```
digi-brain, line 9:
```suggestion
To refer to the actions that we take with units that we transport, we use the terms *embark* and *disembark*.
To refer to the units that we transport, we sometimes use the term *cargo*.
```

modules/special-rules/pages/hitch-x.adoc
```diff
@@ -2,18 +2,18 @@
...
+It can carry as many of these units as we state in brackets.
++
 See the main transportation rules for more about how this works.
 
-NOTE: A {hitch-x} unit does not gain the abilities of embarked units.
-
-.Hitch (X)
+.{hitch-x}
 ====
 The Ork Battlefortress has {hitch-4} -- so you can embark up to four Infantry units in a Battlefortress.
```
IJW-Wartrader, line 18:
```suggestion
The Ork Battlefortress has {hitch-4} -- so you can embark up to four Infantry units on a Battlefortress.
```
Just a very minor suggestion, as the point of Hitch is that they never actually get _in_ the vehicle!

modules/main-rules/pages/transportation.adoc
```diff
@@ -1,59 +1,62 @@
...
-If a {transport-x} unit is destroyed then, just before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the unit suffers a hit.
-If the unit survives, place it in contact with the destroyed {transport-x} unit.
+:!by-hitch-x:
 
-Then remove the {transport-x} unit as normal.
+== Destruction of transportation units
 
-== Transportation by {hitch-x} units
+If a transportation unit is destroyed while cargo units are embarked then, before you remove the unit as a casualty, roll a D6 for each embarked unit: On a result of 1, 2 or 3, the embarked unit suffers a hit.
+If the embarked unit survives, place it completely within 5 cm of the destroyed transportation unit.
```
IJW-Wartrader, line 39: Should this now refer to cargo units throughout? for example:

```suggestion
If a transportation unit is destroyed while cargo units are embarked then, before you remove the unit as a casualty, roll a D6 for each cargo unit: On a result of 1, 2 or 3, the cargo unit suffers a hit.
If the cargo unit survives, place it completely within 5 cm of the destroyed transportation unit.
```

modules/main-rules/pages/transportation.adoc
IJW-Wartrader: I'm a bit confused here, that was the purpose of my note in the first place!

modules/main-rules/pages/transportation.adoc
digi-brain: I started to do exactly that, and then undid it. It would be intelligible either way, and I wasn't 100% sure which is best. But I figured that it is the specific 'embarked' state of the cargo that is most pertinent here.

---

As an aside on the new term:

I added the term 'cargo' in *some* places because there are sentences where it helps to clearly disambiguate the cargo units from other units, especially but not only the transport units.

Part of the trouble is that while 'embarked units' or 'disembarked units' work in many cases, 'unembarked unit' doesn't really seem to fit. And of course 'units that you intend to embark in a moment, but haven't embarked just yet' isn't great either :-D Whereas the term 'cargo unit' seems to work in those cases. We don't really need 'cargo' where 'embarked unit' or 'disembarked unit' cover the situation. And I don't think we really need 'embarked cargo unit' because this is implicit — embarked and disembarked states are relevant only for cargo units.

---

However, we could rationalise it some more, perhaps like this (also breaks down the long sentence that I should've caught before):
```suggestion
If a transportation unit is destroyed then any cargo unit that is embarked at the time may also suffer a hit.
Before you remove the destroyed transport unit, roll a D6 for each embarked unit: On a result of 1, 2 or 3, it suffers a hit.
If the unit survives, place it completely within 5 cm of the destroyed transportation unit.
```

modules/special-rules/pages/hitch-x.adoc
digi-brain: I agree, I'll apply it and resolve.

There's room to quibble over in/on when it comes to classic open-topped battlewagons and the like, but let's not worry about that! :-D

modules/main-rules/pages/transportation.adoc
digi-brain: Basically, my note proposes the opposite of yours — you're working from the assumption that we *have* war engine orders and that those rules are the default, whereas I'm working from opposite assumption, at least for now.
I started this 'conversation' thread so that we don't lose sight of this issue as we near the finish line on this branch (it's much easier to check for unresolved conversations than to hunt for comments on the source, in their various forms), and because we need somewhere to actually *have* the discussion.

IJW-Wartrader: First off, I agree that this needs to be in a thread somewhere, and not just a comment in the source.

But I think we're still saying the same thing in different ways, when I said 'the war engine orders page' I was referring to whatever page discusses how orders interact with war engines. For example the Remastered version of https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/WarEngines/WarEngineMovement.htm

modules/main-rules/pages/transportation.adoc
IJW-Wartrader: Yeah, that was a fairly minor suggestion anyway.

But I agree, the revised version helps break it up a bit more!

IJW-Wartrader:
From my end, this is ready to go, with or without the tweak to destroyed transports.

I've added a note about transported detachment orders to the list of things to deal with in the 'War engine branch plans' issue.

modules/main-rules/pages/transportation.adoc
digi-brain: We're talking at cross purposes, but I think I see where the misunderstanding lies. It sounds like in fact we're actually on the same page. So, how about this:

1. We remove the notes from the code here.
2. We move the empty `war-engine-orders.adoc` placeholder topic from the Special rules section to the Experimental rules section, and update the navigation.
3. We insert a note in `war-engine-orders.adoc`, to remind ourselves to address the point that you raised here. (Just to make sure that we don't overlook it when we get around to that topic.)

In that case we can then tidy up a little here.

However, there is still a fly in the ointment. Although we have this line, which covers detachments of regular units (#52):

>Where transportation units from one detachment carry units from another detachment, both detachments must be on the same orders.

...we don't actually mention what should happen with War Engine transportation units, which don't even *have* orders (yet). I'm inclined to ignore this for now, though I think that we should leave a TODO note as a reminder for us to revisit it at some point. We can also make this note visible so that others are aware. Or you might have an idea for a line we can stick in to clear it up properly now, and then we don't need a note.

```suggestion
{blank}[TODO: We should add something to address what should happen with War Engine transportation units, which don't even have orders (at least, not yet — we intend to experiment with new War Engine orders rules in due course).]
```

What do you think?

digi-brain:
> From my end, this is ready to go, with or without the tweak to destroyed transports.
> 
> I've added a note about transported detachment orders to the list of things to deal with in the 'War engine branch plans' issue.

Okay, cool. I've just added a further query and suggestion, but once we resolve that one way or another then we're good to go I think.

modules/main-rules/pages/transportation.adoc
IJW-Wartrader: > 1. We remove the notes from the code here.
> 2. We move the empty `war-engine-orders.adoc` placeholder topic from the Special rules section to the Experimental rules section, and update the navigation.
> 3. We insert a note in `war-engine-orders.adoc`, to remind ourselves to address the point that you raised here. (Just to make sure that we don't overlook it when we get around to that topic.)

Sounds good to me, and that's all my original note was referring to.


> However, there is still a fly in the ointment. Although we have this line, which covers detachments of regular units (#52):
> 
> > Where transportation units from one detachment carry units from another detachment, both detachments must be on the same orders.
> 
> ...we don't actually mention what should happen with War Engine transportation units, which don't even _have_ orders (yet). I'm inclined to ignore this for now, though I think that we should leave a TODO note as a reminder for us to revisit it at some point. We can also make this note visible so that others are aware. Or you might have an idea for a line we can stick in to clear it up properly now, and then we don't need a note.

OK, in that case we are still talking at cross-purposes. I wouldn't expect any mention of war engines on this page (apart from a Related information link), because I'm assuming that, like the Compendium, all the war engine exceptions are explained in the war engine section, not in the main rules.

digi-brain: >I wouldn't expect any mention of war engines on this page (apart from a Related information link), because ... war engine exceptions are explained in the war engine section

Ah yes, I got side-tracked — one of the perils of getting sucked into particular details that are on a tangent. We are on the same page here.

Though we actually *don't* need the TODO note at `war-engine-orders.adoc` precisely because that *will* inherently integrate with the statement at line #52.

We need the rule/note somewhere else in the war engines section to cover the fact that the standard 'war engines don't have orders' rule *won't* integrate with line #52. I think I'll stick the reminder in `war engines.adoc` for now (alongside another reminder I've already drafted about {no-retreat} and {unbreakable}).

Perhaps in due course we'll want a specific topic of sub-heading like 'War Engine units as transportation'.

```suggestion
```

digi-brain (approved):
Ready to go.

PR #19, opened 2021-08-07
Split speed redux and general movement

IJW-Wartrader:
Lots of pages updated to use split Speed values for all Infantry.

Note that the terrain page will have a conflict for the Dangerous terrain test.


digi-brain (changes requested):
Various suggestions -- some that I'm sure of so I'll apply and resolve them shortly, some that I'll leave so that they are highlighted for further discussion.

modules/main-rules/pages/add-up-assault-values.adoc
```diff
@@ -1,7 +1,8 @@
 = CC1. Add up Assault values
 
 Both players: Add up the Assault values of all your units that are in base-to-base contact with the enemy.
-A maximum of two units can 'gang up' on each enemy unit.
+
+include::partial$ganging-up.adoc[]
```
digi-brain, line 5: The partial makes sense if we need to repeat the note, of course -- but looking at it now, I don't think that this is really a logical place to repeat it anyway. Now that we have it in the proper context at `assault-moves-and-charges.adoc`, I don't think we need to follow GW's original odd logic of having it here.
```suggestion
```

modules/main-rules/pages/orders-and-conditions.adoc
```diff
@@ -30,7 +30,9 @@ Movement phase:: You may make a move with units in this detachment.
 Shooting phase:: You may shoot with the units in this detachment -- but you must *halve* their total Firepower.
 Assault phase:: You *must* move the units in this detachment, at least 5 cm towards the enemy (though not necessarily the nearest enemy).
```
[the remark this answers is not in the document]
IJW-Wartrader: I think this is OK, but I'm slightly wary of the absolute nature of '**must**'.

However that's as much an issue with the `Assault moves and charges` page. What about adding a TIP to that page, to remind players that the requirement to move 5 cm doesn't override failed Leadership tests or Dangerous terrain tests?

Minor suggestion:

```suggestion
Assault phase:: You *must* move the units in this detachment, if they are able to move, and you may initiate a close combat or firefight.
For the full detail, see xref:assault-moves-and-charges.adoc[].
```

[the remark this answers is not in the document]
digi-brain: That's a different point. But I see what you mean — we're setting up a conflict between 'may not move' rules and this 'must move' rule.

Yes, an adjustment to `assault-moves-and-charges.adoc` is best, as that topic should be 'complete' and not have caveats hanging off of it that reside elsewhere. And, in common with my initial suggestion on this thread, I don't think we should *duplicate* it here either.

I'll have a look at `assault-moves-and-charges.adoc` and find a way to tweak it to resolve or clarify the conflict.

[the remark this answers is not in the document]
digi-brain: Added edits to `assault-moves-and-charges.adoc`, so I'll apply the appropriate suggestion here to resolve the particular parts in these lines.

modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -16,9 +16,9 @@ Also, there are some complexities in this Assault phase that mean you need to pa
 When you move a detachment in the Assault phase, move each of its units one after the other.
 For each unit:
 
-* You *must* move your unit at least 5 cm -- or into base contact with an enemy unit that is within 5 cm.
+* Unless your unit is subject to a 'may not move' restriction then you *must* move it at least 5 cm -- or into base contact with an enemy unit that is within 5 cm.
 Any movement beyond this is optional.
-* You *must* move your unit as directly as possible towards an enemy unit.
+* This move *must* be as directly as possible towards an enemy unit.
```
digi-brain, line 21: Hopefully you'll agree that these tweaks resolve the 'may not move versus must move' issue reasonably cleanly.

digi-brain, line 21: Ugh, slight grammar fix required:
```suggestion
* This move *must* be as direct as possible towards an enemy unit.
```

IJW-Wartrader: Where is "a 'may not move' restriction" defined?

digi-brain: I'd argue that we don't need to explicitly define it as it should be reasonably intuitive and the single quote marks hint at this. I did also pre-empt this a bit with some recent tweaks that you may recall, where I switched some other phrases to 'may not move' in the dangerous terrain test and perhaps one or two other places. We can tighten up others as we go along.

IJW-Wartrader: For what it's worth, as a player/reader the phrasing made me wonder where I had missed seeing the definition.

Also the tense is confusing, because a unit that gets stopped by a Dangerous terrain test _becomes_ subject to a 'may not move' condition, rather than _already_ being subject to it before moving.

Obviously it's your call as technical editor, but as a reader it's a sentence that requires a lot of processing, and is ambiguous regarding rules that take effect after the unit starts moving.

digi-brain: Hmm. I don't think it's as bad as that. But I can see that there might be something in what you say. Although, I rather suspect that alternatives may be more convoluted.

I suppose the easiest way out is to use a phrase more along the lines of 'If the unit is able to move', like your suggestion in that other context. I didn't use that because I was trying to be a bit more precise and to mirror the 'may not move' phrasing we've started to add in the actual rules that are relevant for this. But perhaps in this case a more vague appeal for the reader to interpret 'able to move' is actually sufficient?

>and is ambiguous regarding rules that take effect after the unit starts moving

Is it though? I mean, aside from common sense, the 'interrupt' rules themselves should make sense. We're really talking about dangerous terrain tests, and you may recall that we tightened up the phrasing there recently for that exact reason.   Those rules explain what happens with a 'mid-move interrupt' — we shouldn't need to repeat it here.

I'll stew on it some more though; and feel free to suggest ideas. We could insert a NOTE (or a footnote, depending on what seems to be the best fit) if we really have to, but it'd be better to weave it directly into the main text if we can.

IJW-Wartrader: Thinking aloud, some alternatives would be:

You *must*, where possible, move your unit at least 5 cm...
Subject to movement restrictions, you *must* move your unit at least 5 cm...

digi-brain: Yeah, let's keep it simple and go with: `Where possible, you *must* move it at least`...

GitHub won't let me suggest and apply from this specific thread, because "Applying suggestions on deleted lines is not supported". So, I'll resolve this conversation and apply the change on a new one.

modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -16,9 +16,9 @@ Also, there are some complexities in this Assault phase that mean you need to pa
 When you move a detachment in the Assault phase, move each of its units one after the other.
 For each unit:
 
-* You *must* move your unit at least 5 cm -- or into base contact with an enemy unit that is within 5 cm.
+* Unless your unit is subject to a 'may not move' restriction then you *must* move it at least 5 cm -- or into base contact with an enemy unit that is within 5 cm.
```
digi-brain, line 19:
```suggestion
* Where possible, you *must* move it at least 5 cm -- or into base contact with an enemy unit that is within 5 cm.
```

modules/main-rules/pages/terrain-effects-on-movement.adoc
```diff
@@ -31,19 +31,30 @@ NOTE: Hits from dangerous terrain do *not* cause critical damage on War Engine u
 
 == [[roads]]Movement on roads
 
-Where you will move a unit such that its entire move will take place on a road -- in any phase -- add 5 cm to its Speed value for that move.
-Apply this bonus to the unit's base Speed value, before modification.
+Where you will move a unit such that its entire move will take place on a road -- in any phase -- add 5 cm to its Speed value for that move. If the unit has a split Speed value, add 5 cm to its first Speed value and 10 cm to its second Speed value.
```
digi-brain, line 34:
```suggestion
Where you will move a unit such that its entire move will take place on a road -- in any phase -- add 5 cm to its Speed value for that move.
If the unit has a split Speed value, add 5 cm to its first Speed value and 10 cm to its second Speed value.
```

digi-brain (approved):
Good to go I think. Last tiny tweaks resolved, and merge conflicts too.

PR #20, opened 2021-08-07
DC2 units playtest page (Robust)

IJW-Wartrader:


digi-brain:
I thought I'd make this experimental/playtest content my first priority, with the intent of a 'light touch' as previously discussed.

However, I got sucked in and edited rather more than I'd first intended. Anyway, I believe that I have helped with the clarity and consistency -- hopefully I haven't upset things that you find frustrating.

modules/experimental/pages/robust.adoc
```diff
@@ -0,0 +1,102 @@
...
+* When you calculate the detachment strength, count the unit as equal to its remaining Damage Capacity.
+* Each hit reduces its Damage Capacity by one point.
+* If the unit's Damage Capacity is reduced to 0 then it is destroyed.
+
+Where you attack a detachment that includes units that have the {robust} ability:
+
+* Each {robust} unit counts as *two* units for hit allocation.
+This means that when you allocate hit dice and you reach a {robust} unit, you must allocate *two* dice to that unit before you allocate dice to more distant targets.
+// TODO: Do we still have to allocate two dice even if the unit has only 1 DC remaining?
+* Where possible, you must allocate two dice that score equal to or higher than the {robust} unit's Armour value, *or* two dice that score lower.
```
IJW-Wartrader, line 34: > TODO: Do we still have to allocate two dice even if the unit has only 1 DC remaining?

Hmm. I'm not sure. It probably makes sense to change it to the number of remaining DC, but so far it's come up very rarely.

```suggestion
This means that when you allocate hit dice and you reach a {robust} unit, you must allocate as many dice to that unit as it has remaining Damage Capacity, before you allocate dice to more distant targets.
* Where possible, all dice allocated to the {robust} unit must score equal to or higher than its Armour value, *or* all dice allocated must score lower.
```

modules/experimental/pages/robust.adoc
```diff
@@ -0,0 +1,102 @@
...
++
+NOTE: The aim of this restriction is to force the attacker to destroy the unit or bypass it, and so to minimise the need to track units that have only one remaining point of Damage Capacity.
+
+* You may not allocate a third hit die to the unit until you have first allocated a hit die to all eligible units, in the usual way.
+// TODO: Shouldn't this be "a third and fourth hit die" (or "more dice") to be consistent? Or, what if the unit has only 1 DC remaining -- should it then be only "a second die" (or "another die")?
```
IJW-Wartrader, line 39: Agreed.

```suggestion
* You may not allocate a further hit die to the unit until you have first allocated hit dice to all eligible units, in the usual way.
```

modules/experimental/pages/robust.adoc
digi-brain: Okay, allocate against current/remaining DC. That's my gut feeling too.
Your suggestion looks good, I'd just like to tweak it a bit — and we also need to integrate line #31, which isn't included in the selection for this conversation thread. So, I'll make a new suggestion on a new thread that includes that line.

digi-brain, line 34: New suggestion, per earlier thread. Based on yours, but with some tweaks to wording. I also added a little explanation of the last bullet in parentheses at the end — but I'm not wedded to it if you're not keen on that addition.

```suggestion
* Each {robust} unit counts as either *one* or *two* units for hit allocation, according to the number of Damage Capacity points it has left.
This means that when you allocate hit dice and you reach a {robust} unit, you must allocate two dice if it is undamaged or one die if it is damaged.
Then continue to allocate dice to more distant targets in the usual way.
* Where possible, all dice that you allocate to a {robust} unit must score equal to or higher than its Armour value, *or* all dice that you allocate must score lower.
(To minimise the number of damaged units that you need to track in the game.)
```

digi-brain: Oops, didn't need that parenthetical bit as we've covered it in the note in the next line, which is probably a better way to do it anyway :-D

```suggestion
* Each {robust} unit counts as either *one* or *two* units for hit allocation, according to the number of Damage Capacity points it has left.
This means that when you allocate hit dice and you reach a {robust} unit, you must allocate two dice if it is undamaged or one die if it is damaged.
Then continue to allocate dice to more distant targets in the usual way.
* Where possible, all dice that you allocate to a {robust} unit must score equal to or higher than its Armour value, *or* all dice that you allocate must score lower.
```

IJW-Wartrader: Looking good.

digi-brain (approved):
Good to go I reckon. I'll merge it in a moment.

PR #22, opened 2021-08-10
Special abilities continued

### LINKED ISSUE #21: Special abilities continued
This issue is to correlate with the branch `special-abilities-continued`.

As part of other work, we've rounded off content for a lot of the special abilities but there are still a number to complete.

Rather than make my 'first pass' drafts directly on `master` as I did before there were two of us working on the project, it is probably best if all such content starts life on a separate branch.

The scope that I intend for this branch is to populate all of the remaining ability topics and align them with what we've already done.

digi-brain:
Closes #21.


digi-brain:
If you don't have an particular Remastered tweaks in mind for these other special ability topics, then I think that we can proceed to publish this 'first pass' more or less as-is, and leave most of the TODO items for a later date.

Or, you might have some Remastered tweaks in mind that you'd like to thrash out now?

You'll notice that I've moved all of the abilities that originate from 'vehicle variants' to the experimental section, as that was their status in E40k. But I'm sure we'd like to make at least some of them 'official' in due course.

digi-brain:
Actually, I'll just go ahead and publish this. It's just the first pass rewrite of original rules. If we want to improve it or add any Remastered tweaks we can do that at any time.

PR #25, opened 2021-11-25
Special abilities august 2021

IJW-Wartrader:
I think this is ready for checking. In addition to the combat resolution roll additions, I've added a few related information links to assorted pages.

Note that I've changed the titles of CC3 and FF3, but haven't changes the file names as I thought it might be safer to let you do all navigation-level changes.


modules/main-rules/pages/roll-dice-and-add-combat-bonuses.adoc
```diff
@@ -1,6 +1,9 @@
-= CC3. Roll dice and add combat bonuses
+= CC3. Make the combat resolution roll
 
+.Combat resolution roll
 Both players: Roll a D6 and add your close combat bonuses to the result.
 
 * The player with the highest total score wins the combat.
-* If the total scores are the same then the outcome hangs in the balance! Place one Blast marker on every detachment in the combat and then *roll again*.
\ No newline at end of file
+* If the total scores are the same then the outcome hangs in the balance! Place one Blast marker on every detachment in the combat and then *roll again*.
```
digi-brain, line 7: There's an implicit step here, given that the Blast markers in a draw can change the balance of the combat. I think that we should make it explicit:

```suggestion
* If the total scores are the same then the outcome hangs in the balance!
Place one Blast marker on every detachment in the combat, and then proceed again from xref:work-out-close-combat-bonuses.adoc[].
```

IJW-Wartrader: Looks good to me.

modules/main-rules/pages/assault-moves-and-charges.adoc
```diff
@@ -27,6 +27,12 @@ This does not necessarily mean the *closest* enemy unit, or even one that is in
 TIP: You do not have to move units more than 5 cm if you do not want to.
 You might only want to move units into range for a firefight or to grab an objective, for example.
 
+[NOTE]
+.Units that can't move
+====
+The requirement to move at least 5 cm or into base contact does not override any rules that prevent the unit from moving, for example if the detachment fails a Leadership test to move, or the unit fails the first part of a Dangerous terrain test.
+====
+
```
digi-brain, line 35: I think that this part of the topic is getting quite 'noisy' with note blocks. Given that this is really only a clarification that a player will probably only ever need to read once, I think that we could relegate it to a footnote. Possibly also change the relevant line to something like `You *must* move your unit at least 5 cm if you are free to do so -- ...`
I might try that and see how it looks.

modules/main-rules/pages/roll-dice-and-add-firefight-bonuses.adoc
```diff
@@ -1,6 +1,9 @@
-= FF3. Roll dice and add firefight bonuses
+= FF3. Make the combat resolution roll
 
+.Combat resolution roll
 Both players: Roll a D6 and add your firefight bonuses to the result.
 
 * The player with the highest total score wins the firefight.
-* If the total scores are the same then the outcome hangs in the balance! Place one Blast marker on every detachment in the firefight and then *roll again*.
\ No newline at end of file
+* If the total scores are the same then the outcome hangs in the balance! Place one Blast marker on every detachment in the firefight and then *roll again*.
```
digi-brain, line 7: To mirror my edit for the equivalent close combat topic, `roll-dice-and-add-combat-bonuses.adoc`:
```suggestion
* If the total scores are the same then the outcome hangs in the balance!
Place one Blast marker on every detachment in the combat, and then proceed again from xref:work-out-firefight-bonuses.adoc[].
```

modules/main-rules/pages/roll-dice-and-add-firefight-bonuses.adoc
```diff
@@ -1,6 +1,9 @@
-= FF3. Roll dice and add firefight bonuses
+= FF3. Make the combat resolution roll
 
+.Combat resolution roll
```
digi-brain, line 3: Might we need to disambiguate? Perhaps we actually have 'close combat resolution roll' and 'firefight resolution roll'?
Collectively, if/where necessary, 'assault resolution roll'.

IJW-Wartrader: I don't think we gain anything from having two different names. 

We could rename it to 'assault resolution roll', but I suspect that'll just mean that players use a different term than we use in the rules... :-/ I picked 'combat resolution' because that's what everyone seems to use, regardless of game system.

digi-brain: For my liking, 'combat resolution' is much too close to 'close combat resolution' — as distinct from 'firefight resolution'. And we don't call it the 'Combat phase' in this edition, we call it the 'Assault phase'.

I suggest that regardless of what terminology a sample of people may use on casual basis, it pays to be precise and to disambiguate in the actual text. And it's better to be consistent I think — Assault orders, Assault phase, assault move, assault resolution roll.

So I do think that for the general case, 'assault resolution' would be better. And turning my original thought around, we'd use that one term but then qualify it where we need to disambiguate/be specific — 'assault resolution roll for close combat' and 'assault resolution roll for firefight'.

IJW-Wartrader: > For my liking, 'combat resolution' is much too close to 'close combat resolution' — as distinct from 'firefight resolution'. And we don't call it the 'Combat phase' in this edition, we call it the 'Assault phase'.

It's not called the Combat phase or Combat orders in EA either, but the only term I can remember being used across multiple events is 'combat resolution' because that's the standard wargaming term. Note that the original Epic 40k rules use 'the combat' to describe both close combats and firefights.

Personally, I'm very hesitant about choosing terminology that I know is not going to be used by (in my experience) a large proportion of players, but obviously you're the person who has the final say on that.

> And turning my original thought around, we'd use that one term but then qualify it where we need to disambiguate/be specific — 'assault resolution roll for close combat' and 'assault resolution roll for firefight'.

I assume you mean when referring to it in other places, and not within the two resolution pages themselves?

IJW-Wartrader: P.S. My preferred page title would have simply been 'Combat resolution', but of course that then doesn't fit with the rest of the close combat and firefight procedure page titles. :-(

digi-brain: I'm pretty well decided on 'Assault resolution' I think — but I'll try to expound my rationale if only for the sake of posterity:

>Personally, I'm very hesitant about choosing terminology that I know is not going to be used by (in my experience) a large proportion of players, but obviously you're the person who has the final say on that.

I'm not especially, to be honest — people use inaccurate terminology for things all the time, but it doesn't mean that the actual source materials should give up on precision and match colloquial use.

>It's not called the Combat phase or Combat orders in EA either, but the only term I can remember being used across multiple events is 'combat resolution' because that's the standard wargaming term. Note that the original Epic 40k rules use 'the combat' to describe both close combats and firefights.

My point is more that regardless of other editions/games and colloquial terminology, in this edition/game it's called the Assault phase, and the terminology should be consistent with that. The terminology shouldn't wander off and overlap with a sub-category. Terminology isn't as robust as it could be in EA either, but in this case it doesn't present such an issue for one important reason: EA close combats and firefights basically work the same way and share **one procedure**, so it doesn't really matter what you call it — unlike E40k which has a similar but separate procedure for each, and the one can happen after the other.

>I assume you mean when referring to it in other places, and not within the two resolution pages themselves?

Mainly, yes — but when I look again to edit, it might be that the qualifiers are useful to ensure clarity. In fact, see my thoughts on the titles below.

>P.S. My preferred page title would have simply been 'Combat resolution', but of course that then doesn't fit with the rest of the close combat and firefight procedure page titles. :-(

It's good practice to start all 'tasks' in a 'procedure' with a verb. So actually, neither "Combat resolution roll" nor "Combat resolution" are great.

Thus I'm inclined towards something more like "Roll to resolve the assault". Other possibilities include "Roll for assault resolution" or "Make the assault resolution roll", but these are a bit inelegant IMHO.
Or, given that we're talking about two similar but non-identical topics inside two separate procedures, it might pay to include the qualifiers in the titles "Roll to resolve the assault (close combat)" and "Roll to resolve the assault (firefight)". That way we can still refer to "re-roll the roll to resolve the assault" and cover both cases.

modules/special-rules/pages/supreme-commander.adoc
```diff
@@ -9,10 +9,18 @@
...
-If you re-roll the die then you must abide by the second result.
+* If the unit contributes its Assault value to a close combat or contributes its Firepower value to a firefight then you may re-roll a single die when making the combat resolution roll.
+If you re-roll a die then you must abide by the second roll.
 +
-{blank}[TODO: Does "fights in a close combat" include not only base-contact but also 'supporting fire'?]
+//[TODO: Does "fights in a close combat" include not only base-contact but also 'supporting fire'?]
+// IJW editing note: I've changed the text to refer to using the Assault value, based on the original rules's reference to the SC unit itself needing to fight in a CC, not just its bodyguard. I've also tweaked the re-roll text based on the Fate Card Counterattack text, and to allow it to work more easily for players who use highest of 2d6 for combat resolution.
+
+== Tyranid Dominatrix
+In addition to the above, if the Dominatrix supports a close combat you may re-roll a single die when working out who wins the combat.
```
digi-brain, line 19: @IJW-Wartrader I presume you mean to add content here instead of in the `war-engine-supreme-commander.adoc` file, which we can thus delete. I think that's good.

I'd like to approach it slightly differently it terms of layout, and also: Might we want to make this a thing for all cases where a War Engine has the Supreme Commander ability?

```suggestion
Where your unit has the {supreme-commander} ability and is also a War Engine unit, then in addition to the above:

* If the unit adds 'supporting fire' to a close combat, then you may re-roll a single die when work you roll to resolve the close combat. 

[NOTE:e40k]
.War Engine and {supreme-commander}
====
_Epic 40,000_ only states this additional rule in the case of the Tyranid Dominatrix.
We make it a more general rule -- so it also applies to an Ork Warlord mounted in a Gargant, for example.
====
```

modules/special-rules/pages/supreme-commander.adoc
```diff
@@ -9,10 +9,18 @@
 Where your unit has the {supreme-commander} ability:
 
 * All of your detachments that have their HQ unit within 30 cm of the {supreme-commander} unit gain the {stubborn} ability.
-* If this specific unit fights in a close combat or contributes its Firepower to a firefight then you may re-roll the die that decides the outcome of that close combat or firelight.
-If you re-roll the die then you must abide by the second result.
+* If the unit contributes its Assault value to a close combat or contributes its Firepower value to a firefight then you may re-roll a single die when making the combat resolution roll.
+If you re-roll a die then you must abide by the second roll.
```
digi-brain, line 13: I think it'd still make best sense to stick with singular here, because that is the default and the optional alternate method explains itself anyway.
Also, I intend to separate 'close combat' and 'firefight' clearly throughout, and try "Assault resolution" as the more general term. So, I put it that way in this suggestion and I'll tweak others in similar fashion and see how it looks overall.

```suggestion
* If the unit contributes its Assault value to a close combat, or contributes its Firepower value to a firefight, then you may re-roll the die for Assault resolution.
If you re-roll the die then you must abide by the second roll.
```

modules/main-rules/pages/roll-dice-and-add-firefight-bonuses.adoc
digi-brain: This conversation is now outdated by various changes, but I'll keep an eye open for other instances where terminology needs to change to be consistent.

modules/main-rules/pages/assault-moves-and-charges.adoc
digi-brain: This conversation is now outdated by change in which I've implemented the footnote approach. So I'll resolve.

modules/special-rules/pages/psyker.adoc
```diff
@@ -16,6 +16,7 @@ Where your unit has the {psyker} ability:
 * You may gain an advantage in a close combat or firefight if this unit is involved.
 For more about this, see xref:main-rules:work-out-close-combat-bonuses.adoc[] and xref:main-rules:work-out-firefight-bonuses.adoc[].
 // TODO: Do we need to define "is involved", either here or in the actual bonus tables? The text in the tables is a bit vague and only loosely correlates with the rules text -- "fighting" versus "in base-to-base contact", "supporting" rather than "within 15 cm" (though I suppose "most psykers in the firefight" works okay given the way that the trigger for a firefight is defined).
+// IJW editing note: I've added a new conditional footnote to the assault bonuses table.
```
digi-brain, line 19: Now redundant. Also note that I've reworked the assault bonuses table to streamline it a bit and make it work for the quick reference too -- your 'conditional' point is still in there of course, just a slightly different way.
```suggestion
```

modules/special-rules/pages/supreme-commander.adoc
IJW-Wartrader: > I'd like to approach it slightly differently it terms of layout, and also: Might we want to make this a thing for all cases where a War Engine has the Supreme Commander ability?

I suggested that on FB a while ago, and it was... _unpopular_...

digi-brain: Ah, okay. In that case, something like this:

```suggestion
[NOTE]
.Tyranid Dominatrix -- additional effect for {supreme-commander}
====
The Tyranid Dominatrix unit is a special case.
Not only does it have the {supreme-commander} ability, but for this unit only the ability has this additional effect:

* If the unit adds 'supporting fire' to a close combat, then you may re-roll a single die when work you roll to resolve the close combat. 
====
```
It makes rather an odd exception of course. Given that it is specific to one unit, it'd probably be best to move this rule to the actual Tyranid Dominatrix datasheet when we get to it.

**Aside**: We may also need to tighten up use of terms like 'supporting fire' in due course, to be 100% clear about when we mean that a rule applies where the unit supports a close combat where its own detachment is engaged (our main use of 'supporting fire'), and when we mean that a rule does or does not apply where the unit 'supports' a friendly detachment in a separate close combat. But I think that we leave this as a discrete task to tackle later.

IJW-Wartrader: > It makes rather an odd exception of course. Given that it is specific to one unit, it'd probably be best to move this rule to the actual Tyranid Dominatrix datasheet when we get to it.

I think it still needs to be in the rules as well, I just didn't want to add an extra ability and page for something that only affects one single unit. :-(

> **Aside**: We may also need to tighten up use of terms like 'supporting fire' in due course, to be 100% clear about when we mean that a rule applies where the unit supports a close combat where its own detachment is engaged (our main use of 'supporting fire'), and when we mean that a rule does or does not apply where the unit 'supports' a friendly detachment in a separate close combat. But I think that we leave this as a discrete task to tackle later.

Agreed, but in this case the term we're currently using in the War Engine section is 'support', rather than 'supporting fire'.

Would this work?
```suggestion
[NOTE]
.Tyranid Dominatrix -- additional effect for {supreme-commander}
====
The Tyranid Dominatrix unit is a special case.
Not only does it have the {supreme-commander} ability, but for this unit only the ability has this additional effect:

* If the unit supports another detachment that is in close combat, then you may re-roll a single die when work you roll to resolve the close combat. 
====
```

digi-brain: I'd misapprehended a point I won't bore you with, but I can see where I'd gone wrong now. Yes, given the actual situation and given our current use of terms, I agree that we should not use 'supporting fire' in this case.

I've though of some further tweaks though -- in part to clarify some more, and in part to mirror the latest language in the earlier lines.

```suggestion
[NOTE]
.Tyranid Dominatrix -- additional effect for {supreme-commander}
====
The Tyranid Dominatrix unit is a special case.
Not only is it a War Engine that has the {supreme-commander} ability, but for this unit only the ability has this additional effect:

* If the Dominatrix supports another detachment that is in close combat, then you may re-roll the die for Assault Resolution in that close combat.
If you re-roll the die then you must abide by the second roll.
====
```

IJW-Wartrader: Looks good to me!

digi-brain: Cool. I'll commit this, and make time soon to build and make final checks with a view to approve/merge/publish. :)

modules/special-rules/pages/supreme-commander.adoc
```diff
@@ -9,10 +9,25 @@
...
+[NOTE]
+.Tyranid Dominatrix -- additional effect for {supreme-commander}
+====
+The Tyranid Dominatrix unit is a special case.
+Not only is it a War Engine that has the {supreme-commander} ability, but for this unit only the ability has this additional effect:
+
+* If the Dominatrix supports another detachment that is in close combat, then you may re-roll the die for Assault Resolution in that close combat.
+If you re-roll the die then you must abide by the second roll.
+====
+
```
digi-brain, line 27:
```suggestion
```

digi-brain (approved):
Ready to go. I also rationalised various bits of related assault rules texts, just to improve general phrasing and consistency — nothing important. I've spotted other CC/FF stuff that could be a bit more consistent, but again, only 'nice to have' and can wait for another time.

PR #28, opened 2021-12-01
War Engines first pass

IJW-Wartrader:
Mostly related to hit allocation.

### DIFF
--- a/modules/main-rules/pages/attempt-repairs.adoc
+++ b/modules/main-rules/pages/attempt-repairs.adoc
@@ -2,10 +2,10 @@
 
 This step is relevant only for war engines with minor (repairable) damage or collapsed void shields.
 
-For each repairable critical result and each collapsed shield, take a Repair test.
+For each repairable critical damage effect and each collapsed void shield, take a Repair test.
 
 .Repair test
-Roll a D6: On a score of 4 or more you succeed -- the critical result or void shield is repaired.
+Roll a D6: On a score of 4 or more you succeed -- the critical damage effect or void shield is repaired.
 
 .Related information
 * xref:special-rules:war-engine-critical-damage.adoc[]
--- a/modules/main-rules/pages/attempt-repairs.adoc
+++ b/modules/main-rules/pages/attempt-repairs.adoc
@@ -1,11 +1,11 @@
 = R3. Attempt repairs
 
-This step is relevant only for war engines with minor (repairable) damage or collapsed shields.
+This step is relevant only for war engines with minor (repairable) damage or collapsed void shields.
 
-For each damaged location and each collapsed shield, take a Repair test.
+For each repairable critical result and each collapsed shield, take a Repair test.
 
 .Repair test
-Roll a D6: On a score of 4 or more you succeed -- the location or shield is repaired.
+Roll a D6: On a score of 4 or more you succeed -- the critical result or void shield is repaired.
 
 .Related information
 * xref:special-rules:war-engine-critical-damage.adoc[]
--- a/modules/main-rules/pages/roll-for-close-combat-hits.adoc
+++ b/modules/main-rules/pages/roll-for-close-combat-hits.adoc
@@ -4,8 +4,9 @@ In close combat Armour values are ignored and instead you get a hit if you equal
 Look up the combat difference on the table to find out the hit numbers for winning and losing units.
 
 Roll a D6 for each of your units in base-to-base contact with the enemy, even if you lost.
-Don't forget that a maximum of two units can attack one enemy unit in close combat.
-You can choose which enemy units to apply hits against but as with shooting you can only put a second hit on a unit once all of the enemy units have been allocated one hit each.
+Don't forget that you can place a maximum of two units in base contact with an enemy unit.
+You can choose which enemy units to apply hits against but as with shooting you can only put a second hit on a unit once all of the enemy units in base contact have been allocated one hit each.
+// IJW editing note: tweaked to use more consistent references to base contact.
 
 include::partial$close-combat-results-table.adoc[]
 
--- a/modules/main-rules/pages/roll-for-close-combat-hits.adoc
+++ b/modules/main-rules/pages/roll-for-close-combat-hits.adoc
@@ -5,8 +5,7 @@ Look up the combat difference on the table to find out the hit numbers for winni
 
 Roll a D6 for each of your units in base-to-base contact with the enemy, even if you lost.
 Don't forget that you can place a maximum of two units in base contact with an enemy unit.
-You can choose which enemy units to apply hits against but as with shooting you can only put a second hit on a unit once all of the enemy units in base contact have been allocated one hit each.
-// IJW editing note: tweaked to use more consistent references to base contact.
+You can choose which enemy units to apply hits against -- but, similar to the allocation of shooting hits, you may not allocate a second hit to a unit unless you have already allocated a hit to every unit that is in base contact.
 
 include::partial$close-combat-results-table.adoc[]
 
--- a/modules/special-rules/pages/shooting-at-war-engines.adoc
+++ b/modules/special-rules/pages/shooting-at-war-engines.adoc
@@ -9,11 +9,8 @@ These categories reflect the fact that these units are so large and easy to hit.
 
 * A War Engine unit can become Immobilised -- and so even easier to hit -- as a result of *critical damage*.
 * There is no target category for 'Immobilised War Engine in the open' -- instead, you simply roll as many hit dice as the total Firepower that you direct at the unit.
-
-{blank}[TODO: Cross-reference the Firepower table, once we've added it. Also, we actually target detachments not units, and we need to clarify what happens when a War Engine detachment has a mix of Immobilised and non-Immobilised WE units.]
-{blank}[TODO: reply IJW, I think the easiest is to treat it as a detachment that's partially in cover - you can concentrate on the IMM units and get more dice, or potentially hit everyone and get fewer dice.]
-
-If a detachment of War Engine units has a mix of mobile and immobilised units, then you may choose to shoot at only the units that are immobilised, if you wish. If you do this then, when you allocate hits later in this procedure, you may not allocate hits to any of the units that still mobile.
+* If a target detachment contains both mobile and Immobilised War Engine units, then you may choose to shoot at only those that are Immobilised, if you wish.
+If you do this then you may allocate hits to the Immobilised units **only**.
 
 == War Engine units and Damage Capacity
 
@@ -32,11 +29,8 @@ Or, you may like to use coloured dice or 'damage tracker' dials -- when a War En
 
 == Allocation of hits in War Engine detachments
 
-Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way. So allocate one hit to the nearest available target, then the next closest, and so on.
-// IJW editing note, I've tweaked this by copying text from S6 Allocate hits, to reinforce that you do the standard front-to-back hit allocation, one hit at a time.
-
-{blank}[TODO: Clarify whether we allocate against initial or current Damage Capacity. Current, I presume?]
-{blank}[TODO: reply IJW, as originally written, neither. You allocate as if they were regular units.]
+Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way.
+So, allocate one hit to the nearest available target unit, then the next nearest, and so on.
 
 .Related information
 * xref:war-engine-critical-damage.adoc[]
--- a/modules/special-rules/pages/shooting-at-war-engines.adoc
+++ b/modules/special-rules/pages/shooting-at-war-engines.adoc
@@ -10,7 +10,7 @@ These categories reflect the fact that these units are so large and easy to hit.
 * A War Engine unit can become Immobilised -- and so even easier to hit -- as a result of *critical damage*.
 * There is no target category for 'Immobilised War Engine in the open' -- instead, you simply roll as many hit dice as the total Firepower that you direct at the unit.
 * If a target detachment contains both mobile and Immobilised War Engine units, then you may choose to shoot at only those that are Immobilised, if you wish.
-If you do this then you may allocate hits to the Immobilised units **only**.
+If you do this then you may allocate hits to the Immobilised units *only*.
 
 == War Engine units and Damage Capacity
 
--- a/modules/special-rules/pages/shooting-at-war-engines.adoc
+++ b/modules/special-rules/pages/shooting-at-war-engines.adoc
@@ -1,6 +1,6 @@
 = Shooting at war engines
 
-Resolve shooting at War Engine detachments in much the same way as for other detachments -- there are just a few more details to consider. 
+Resolve shooting at War Engine detachments in much the same way as for other detachments -- there are just a few more details to consider.
 
 == War Engine detachments and the Firepower table
 
@@ -11,6 +11,9 @@ These categories reflect the fact that these units are so large and easy to hit.
 * There is no target category for 'Immobilised War Engine in the open' -- instead, you simply roll as many hit dice as the total Firepower that you direct at the unit.
 
 {blank}[TODO: Cross-reference the Firepower table, once we've added it. Also, we actually target detachments not units, and we need to clarify what happens when a War Engine detachment has a mix of Immobilised and non-Immobilised WE units.]
+{blank}[TODO: reply IJW, I think the easiest is to treat it as a detachment that's partially in cover - you can concentrate on the IMM units and get more dice, or potentially hit everyone and get fewer dice.]
+
+If a detachment of War Engine units has a mix of mobile and immobilised units, then you may choose to shoot at only the units that are immobilised, if you wish. If you do this then, when you allocate hits later in this procedure, you may not allocate hits to any of the units that still mobile.
 
 == War Engine units and Damage Capacity
 
@@ -29,9 +32,11 @@ Or, you may like to use coloured dice or 'damage tracker' dials -- when a War En
 
 == Allocation of hits in War Engine detachments
 
-Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way.
-In other words, you must allocate hits evenly among the target War Engine units that are within the range and line of fire of your attacking units.
+Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way. So allocate one hit to the nearest available target, then the next closest, and so on.
+// IJW editing note, I've tweaked this by copying text from S6 Allocate hits, to reinforce that you do the standard front-to-back hit allocation, one hit at a time.
+
 {blank}[TODO: Clarify whether we allocate against initial or current Damage Capacity. Current, I presume?]
+{blank}[TODO: reply IJW, as originally written, neither. You allocate as if they were regular units.]
 
 .Related information
 * xref:war-engine-critical-damage.adoc[]
--- a/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
+++ b/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
@@ -7,6 +7,3 @@ Or, it may be completely annihilated in a spectacular fashion that presents a ha
 
 When a War Engine unit's current Damage Capacity value reaches `0`, it suffers *catastrophic damage*.
 Roll a D6 and check the result against the Catastrophic Damage Table on that unit's data sheet.
-
-{blank}[TODO: Ideally, we should change some terminology to disambiguate the general usage of the term 'destroyed' from the specific result of 'Destroyed' (rather than 'Wrecked') on catastrophic damage tables.]
-// IJW reply: Biotitans are 'Killed' rather than 'Wrecked'. I suggest 'Annihilated' to replace 'Destroyed' in the results.
--- a/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
+++ b/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
@@ -1,8 +1,7 @@
 = War engine catastrophic damage
 
-A war engine can be 'destroyed' to different degrees.
-It may be only wrecked -- out of the battle, but perhaps not out of the war.
-Or, it may be truly destroyed in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
+A destroyed war engine may be only wrecked or killed -- out of the battle, but possibly salvageable.
+Or, it may be completely annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
 
 ---
 
@@ -10,3 +9,4 @@ When a War Engine unit's current Damage Capacity value reaches `0`, it suffers *
 Roll a D6 and check the result against the Catastrophic Damage Table on that unit's data sheet.
 
 {blank}[TODO: Ideally, we should change some terminology to disambiguate the general usage of the term 'destroyed' from the specific result of 'Destroyed' (rather than 'Wrecked') on catastrophic damage tables.]
+// IJW reply: Biotitans are 'Killed' rather than 'Wrecked'. I suggest 'Annihilated' to replace 'Destroyed' in the results.
--- a/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
+++ b/modules/special-rules/pages/war-engine-catastrophic-damage.adoc
@@ -1,7 +1,7 @@
 = War engine catastrophic damage
 
-A destroyed war engine may be only wrecked or killed -- out of the battle, but possibly salvageable.
-Or, it may be completely annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
+A destroyed war engine might be only wrecked.
+Or, it might be annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
 
 ---
 
--- a/modules/special-rules/pages/war-engine-critical-damage.adoc
+++ b/modules/special-rules/pages/war-engine-critical-damage.adoc
@@ -27,7 +27,7 @@ Some results indicate that the War Engine is Immobilised, and thus unable to mov
 Mark this as appropriate and apply the rules for the Immobilised condition while it remains in effect.
 
 Some results indicate that the effect only lasts 'until repaired'.
-You may attempt to repair each repairable critical result on your War Engine at the appropriate step in the Rally phase.
+You may attempt to repair each repairable critical damage effect on your War Engine at the appropriate step in the Rally phase.
 
 .Related information
 * xref:main-rules:orders-and-conditions.adoc#Immobilised[Immobilised]
--- a/modules/special-rules/pages/war-engine-critical-damage.adoc
+++ b/modules/special-rules/pages/war-engine-critical-damage.adoc
@@ -17,7 +17,6 @@ There is no further effect.
 Make a 2D6 roll on the Critical Damage Table on the War Engine unit's data sheet, to find out what happens.
 
 The Critical Damage Table describes what additional effect the hit has.
-// IJW editing note: I've removed the references to locations, as a number of critical results (notable 'Fire' on Gargants and others, and Stunned, Deep Wound etc. on Biotitans) are not associated with a specific location.
 Critical damage takes many forms -- inoperable weapons, raging fires, crippled engines, and so on.
 The table also states how many extra points to deduct from the War Engine's current Damage Capacity.
 
--- a/modules/special-rules/pages/war-engine-critical-damage.adoc
+++ b/modules/special-rules/pages/war-engine-critical-damage.adoc
@@ -16,7 +16,8 @@ There is no further effect.
 * *6*: The hit causes critical damage.
 Make a 2D6 roll on the Critical Damage Table on the War Engine unit's data sheet, to find out what happens.
 
-The Critical Damage Table describes what critical location is damaged by the hit, and it details the effects on the systems at that location.
+The Critical Damage Table describes what additional effect the hit has.
+// IJW editing note: I've removed the references to locations, as a number of critical results (notable 'Fire' on Gargants and others, and Stunned, Deep Wound etc. on Biotitans) are not associated with a specific location.
 Critical damage takes many forms -- inoperable weapons, raging fires, crippled engines, and so on.
 The table also states how many extra points to deduct from the War Engine's current Damage Capacity.
 
@@ -26,7 +27,7 @@ Some results indicate that the War Engine is Immobilised, and thus unable to mov
 Mark this as appropriate and apply the rules for the Immobilised condition while it remains in effect.
 
 Some results indicate that the effect only lasts 'until repaired'.
-You may attempt to repair each damaged location on your War Engine at the appropriate step in the Rally phase.
+You may attempt to repair each repairable critical result on your War Engine at the appropriate step in the Rally phase.
 
 .Related information
 * xref:main-rules:orders-and-conditions.adoc#Immobilised[Immobilised]
--- a/modules/special-rules/pages/war-engine-movement.adoc
+++ b/modules/special-rules/pages/war-engine-movement.adoc
@@ -7,9 +7,10 @@ The sheer mass of a war engine makes it lumbering and slow to manoeuvre.
 == Forward movement of War Engine units
 
 In general, when you move a War Engine unit, you may only move it in a straight line, directly ahead.
-You may make a single turn of up to 45° for free as part of the move.
-// IJW editing note: I've tweaked this to make it clearer that you get a single free turn, and not multiple free turns that add up to 45º. The original GW text was ambiguous, but the example made it clear that they _meant_ one turn...
-You may make more turns, at a cost of 5 cm of the unit's move per turn of up to 45°.
+However:
+
+* You may make a single turn of up to 45° for free as part of the move.
+* You may make more turns, at a cost of 5 cm of the unit's move per extra turn of up to 45°.
 
 .Forward movement of War Engine units
 ====
--- a/modules/special-rules/pages/war-engine-movement.adoc
+++ b/modules/special-rules/pages/war-engine-movement.adoc
@@ -33,7 +33,7 @@ But then you would have to take your chances with a dangerous terrain test.
 == Reverse movement of War Engine units
 
 You may move a War Engine unit backwards, though not as fast.
-Make the move in the same way as for forward movement, except that you cannot move the unit more than 5 cm.
+Make the move in the same way as for forward movement, except that you may not move the unit more than 5 cm.
 
 == Movement of War Engine units in the Assault phase
 
--- a/modules/special-rules/pages/war-engine-movement.adoc
+++ b/modules/special-rules/pages/war-engine-movement.adoc
@@ -7,7 +7,8 @@ The sheer mass of a war engine makes it lumbering and slow to manoeuvre.
 == Forward movement of War Engine units
 
 In general, when you move a War Engine unit, you may only move it in a straight line, directly ahead.
-You may turn the unit by up to 45° for free as part of the move.
+You may make a single turn of up to 45° for free as part of the move.
+// IJW editing note: I've tweaked this to make it clearer that you get a single free turn, and not multiple free turns that add up to 45º. The original GW text was ambiguous, but the example made it clear that they _meant_ one turn...
 You may make more turns, at a cost of 5 cm of the unit's move per turn of up to 45°.
 
 .Forward movement of War Engine units
@@ -32,13 +33,12 @@ But then you would have to take your chances with a dangerous terrain test.
 == Reverse movement of War Engine units
 
 You may move a War Engine unit backwards, though not as fast.
-Make the move in the same way as for forward movement, except that you must pay an additional 5 cm of the unit's movement.
-{blank}[TODO: Is a mere 5 cm reduction for a reverse move too liberal, should it be a bit more restrictive than this? Also, should we allow reverse movement for all war engines, or only for a sub-set such as super-heavy tanks (as implied in the original note from Firepower #1)?]
+Make the move in the same way as for forward movement, except that you cannot move the unit more than 5 cm.
 
 == Movement of War Engine units in the Assault phase
 
 In the Assault phase, you may choose *not* to move units in a War Engine detachment -- but then you may not initiate close combats or firefights with them either.
-Otherwise, War Engine units are subject to the same movement restrictions in the Assault phase as other units.
+Otherwise, War Engine units are subject to the same movement restrictions in the Assault phase as other units, but their own movement restrictions take priority.
 
 == War engine units and dangerous terrain tests
 
--- a/modules/special-rules/pages/war-engine-shields.adoc
+++ b/modules/special-rules/pages/war-engine-shields.adoc
@@ -25,11 +25,8 @@ You can track remaining fields/shields in a similar way to Damage Capacity -- ju
 
 The main difference between the two technologies is that on collapse, power fields are destroyed and of no further use in the battle, while void shields may be repaired during the battle to provide protection again.
 You check to see if void shields are repaired at the appropriate step in the Rally phase.
-// IJW editing note: I've changed all 'recover' references to 'repair' for consistency with other pages.
 
 Neither power fields nor void shields offer any protection from close combat hits, as the enemy is inside the protective envelope at that point.
-{blank}[TODO: " -- though they do protect in a firefight" perhaps?]
-// IJW reply: I don't think the firefight text is needed, as this is effectively a duplicate of the CC page text, and there is no equivalent text on the firefight page.
 
 .Related information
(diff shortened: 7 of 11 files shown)


modules/main-rules/pages/attempt-repairs.adoc
```diff
@@ -1,11 +1,11 @@
 = R3. Attempt repairs
 
-This step is relevant only for war engines with minor (repairable) damage or collapsed shields.
+This step is relevant only for war engines with minor (repairable) damage or collapsed void shields.
 
-For each damaged location and each collapsed shield, take a Repair test.
+For each repairable critical result and each collapsed shield, take a Repair test.
```
IJW-Wartrader, line 5:
```suggestion
For each repairable critical result and each collapsed void shield, take a Repair test.
```

Just spotted a 'void' that I missed.

digi-brain: Very slight additional tweaks to be semantically precise.

```suggestion
For each repairable critical damage effect and each collapsed void shield, take a Repair test.
```

modules/main-rules/pages/attempt-repairs.adoc
```diff
@@ -1,11 +1,11 @@
...
 
 .Repair test
-Roll a D6: On a score of 4 or more you succeed -- the location or shield is repaired.
+Roll a D6: On a score of 4 or more you succeed -- the critical result or void shield is repaired.
```
digi-brain, line 8:
```suggestion
Roll a D6: On a score of 4 or more you succeed -- the critical damage effect or void shield is repaired.
```

modules/main-rules/pages/roll-for-close-combat-hits.adoc
```diff
@@ -4,8 +4,9 @@ In close combat Armour values are ignored and instead you get a hit if you equal
 Look up the combat difference on the table to find out the hit numbers for winning and losing units.
 
 Roll a D6 for each of your units in base-to-base contact with the enemy, even if you lost.
-Don't forget that a maximum of two units can attack one enemy unit in close combat.
-You can choose which enemy units to apply hits against but as with shooting you can only put a second hit on a unit once all of the enemy units have been allocated one hit each.
+Don't forget that you can place a maximum of two units in base contact with an enemy unit.
+You can choose which enemy units to apply hits against but as with shooting you can only put a second hit on a unit once all of the enemy units in base contact have been allocated one hit each.
+// IJW editing note: tweaked to use more consistent references to base contact.
```
digi-brain, line 9: Yup, seems good to me. I was just going to remove the editing note and pass that straight through, and then I spotted some other minor issues in the rest of this bit of text. I suggest his instead.

(Basically I've eliminated passive phrasing, added some punctuation to break up the long sentence a bit, and rephrased because the allocation of shooting hits isn't _quite_ the same because it depends on 'closeness' rather than 'base contact'.) 

```suggestion
Don't forget that you can place a maximum of two units in base contact with an enemy unit.
You can choose which enemy units to apply hits against -- but, similar to the allocation of shooting hits, you may not allocate a second hit to a unit unless you have already allocated a hit to every unit that is in base contact.
// IJW editing note: tweaked to use more consistent references to base contact.
```

modules/special-rules/pages/shooting-at-war-engines.adoc
```diff
@@ -11,6 +11,9 @@ These categories reflect the fact that these units are so large and easy to hit.
 * There is no target category for 'Immobilised War Engine in the open' -- instead, you simply roll as many hit dice as the total Firepower that you direct at the unit.
 
 {blank}[TODO: Cross-reference the Firepower table, once we've added it. Also, we actually target detachments not units, and we need to clarify what happens when a War Engine detachment has a mix of Immobilised and non-Immobilised WE units.]
+{blank}[TODO: reply IJW, I think the easiest is to treat it as a detachment that's partially in cover - you can concentrate on the IMM units and get more dice, or potentially hit everyone and get fewer dice.]
+
+If a detachment of War Engine units has a mix of mobile and immobilised units, then you may choose to shoot at only the units that are immobilised, if you wish. If you do this then, when you allocate hits later in this procedure, you may not allocate hits to any of the units that still mobile.
```
digi-brain, line 16: Good solution, makes sense to mirror the cover/not in cover situation. Just some tidy up to do, and a little rephrasing.

```suggestion
* If a target detachment contains both mobile and Immobilised War Engine units, then you may choose to shoot at only those that are Immobilised, if you wish.
If you do this then you may allocate hits to the Immobilised units **only**.
```

modules/special-rules/pages/shooting-at-war-engines.adoc
```diff
@@ -29,9 +32,11 @@ Or, you may like to use coloured dice or 'damage tracker' dials -- when a War En
 
 == Allocation of hits in War Engine detachments
 
-Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way.
-In other words, you must allocate hits evenly among the target War Engine units that are within the range and line of fire of your attacking units.
+Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way. So allocate one hit to the nearest available target, then the next closest, and so on.
+// IJW editing note, I've tweaked this by copying text from S6 Allocate hits, to reinforce that you do the standard front-to-back hit allocation, one hit at a time.
+
 {blank}[TODO: Clarify whether we allocate against initial or current Damage Capacity. Current, I presume?]
+{blank}[TODO: reply IJW, as originally written, neither. You allocate as if they were regular units.]
```
digi-brain, line 39:
```suggestion
Where the target detachment consists of more than one War Engine unit, allocate hits in the usual way.
So, allocate one hit to the nearest available target unit, then the next nearest, and so on.
```

Note to self: We have a mix of 'closest' and 'nearest' at _S6 Allocate hits_, even in the same sentence. It's minor, but might as well update it to use a single term throughout the topic, and to match the edit I've made in this suggestion (that is, use 'nearest').

modules/main-rules/pages/attempt-repairs.adoc
IJW-Wartrader: Looks good.

modules/main-rules/pages/attempt-repairs.adoc
IJW-Wartrader: Also looks good.

modules/main-rules/pages/roll-for-close-combat-hits.adoc
IJW-Wartrader: Maybe take out the editing note too, as it will be obsolete.

modules/special-rules/pages/shooting-at-war-engines.adoc
IJW-Wartrader: For what it's worth, this was adapted directly from work-out-how-many-hit-dice-to-roll.adoc:

> If the target detachment has some units in cover and some in the open, then you may choose to shoot at only the units that are the open, if you wish. If you do this then, when you allocate hits later in this procedure, you may not allocate hits to any of the units that are in cover.

modules/special-rules/pages/shooting-at-war-engines.adoc
IJW-Wartrader: Aye and aye.

modules/special-rules/pages/war-engine-catastrophic-damage.adoc
```diff
@@ -1,12 +1,12 @@
 = War engine catastrophic damage
 
-A war engine can be 'destroyed' to different degrees.
-It may be only wrecked -- out of the battle, but perhaps not out of the war.
-Or, it may be truly destroyed in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
+A destroyed war engine may be only wrecked or killed -- out of the battle, but possibly salvageable.
+Or, it may be completely annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
```
digi-brain, line 4: Looks good. I just have some tweaks for a variety of minor reasons grammatical or semantic reasons. I was going to suggest this:

```suggestion
A destroyed war engine or bio-construct might be only wrecked or killed -- out of the battle, but potentially salvageable.
Or, it may be annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
```
...but it occurs to me that you can't really 'salvage' a 'killed' bio-construct...! A little more thought required I think.

modules/special-rules/pages/war-engine-catastrophic-damage.adoc
```diff
@@ -1,12 +1,12 @@
...
 
 ---
 
 When a War Engine unit's current Damage Capacity value reaches `0`, it suffers *catastrophic damage*.
 Roll a D6 and check the result against the Catastrophic Damage Table on that unit's data sheet.
 
-{blank}[TODO: Ideally, we should change some terminology to disambiguate the general usage of the term 'destroyed' from the specific result of 'Destroyed' (rather than 'Wrecked') on catastrophic damage tables.]
\ No newline at end of file
+{blank}[TODO: Ideally, we should change some terminology to disambiguate the general usage of the term 'destroyed' from the specific result of 'Destroyed' (rather than 'Wrecked') on catastrophic damage tables.]
+// IJW reply: Biotitans are 'Killed' rather than 'Wrecked'. I suggest 'Annihilated' to replace 'Destroyed' in the results.
```
digi-brain, line 12: Yes, I agree — 'Annihilated' disambiguates well so that we have no overlap with the more usual use of 'destroyed'.
```suggestion
```

modules/special-rules/pages/war-engine-critical-damage.adoc
```diff
@@ -16,7 +16,8 @@ There is no further effect.
 * *6*: The hit causes critical damage.
 Make a 2D6 roll on the Critical Damage Table on the War Engine unit's data sheet, to find out what happens.
 
-The Critical Damage Table describes what critical location is damaged by the hit, and it details the effects on the systems at that location.
+The Critical Damage Table describes what additional effect the hit has.
+// IJW editing note: I've removed the references to locations, as a number of critical results (notable 'Fire' on Gargants and others, and Stunned, Deep Wound etc. on Biotitans) are not associated with a specific location.
```
digi-brain, line 20: Like it — more universal, and more concise. Just tidy-up to do here.
```suggestion
The Critical Damage Table describes what additional effect the hit has.
```

modules/special-rules/pages/war-engine-critical-damage.adoc
```diff
@@ -26,7 +27,7 @@ Some results indicate that the War Engine is Immobilised, and thus unable to mov
 Mark this as appropriate and apply the rules for the Immobilised condition while it remains in effect.
 
 Some results indicate that the effect only lasts 'until repaired'.
-You may attempt to repair each damaged location on your War Engine at the appropriate step in the Rally phase.
+You may attempt to repair each repairable critical result on your War Engine at the appropriate step in the Rally phase.
```
digi-brain, line 30:
```suggestion
You may attempt to repair each repairable critical damage effect on your War Engine at the appropriate step in the Rally phase.
```

modules/special-rules/pages/war-engine-movement.adoc
```diff
@@ -7,7 +7,8 @@ The sheer mass of a war engine makes it lumbering and slow to manoeuvre.
 == Forward movement of War Engine units
 
 In general, when you move a War Engine unit, you may only move it in a straight line, directly ahead.
-You may turn the unit by up to 45° for free as part of the move.
+You may make a single turn of up to 45° for free as part of the move.
+// IJW editing note: I've tweaked this to make it clearer that you get a single free turn, and not multiple free turns that add up to 45º. The original GW text was ambiguous, but the example made it clear that they _meant_ one turn...
```
digi-brain, line 11: Yes, I agree, good catch.
```suggestion
You may make a single turn of up to 45° for free as part of the move.
```

modules/special-rules/pages/war-engine-movement.adoc
```diff
@@ -32,13 +33,12 @@ But then you would have to take your chances with a dangerous terrain test.
 == Reverse movement of War Engine units
 
 You may move a War Engine unit backwards, though not as fast.
-Make the move in the same way as for forward movement, except that you must pay an additional 5 cm of the unit's movement.
-{blank}[TODO: Is a mere 5 cm reduction for a reverse move too liberal, should it be a bit more restrictive than this? Also, should we allow reverse movement for all war engines, or only for a sub-set such as super-heavy tanks (as implied in the original note from Firepower #1)?]
+Make the move in the same way as for forward movement, except that you cannot move the unit more than 5 cm.
```
digi-brain, line 36: Good — just a minor tweak to phrasing.
```suggestion
Make the move in the same way as for forward movement, except that you may not move the unit more than 5 cm.
```

modules/special-rules/pages/war-engine-shields.adoc
```diff
@@ -23,11 +23,13 @@ When all of the unit's fields/shields are collapsed then any further hits will c
 You can track remaining fields/shields in a similar way to Damage Capacity -- just be sure to differentiate so that you are clear about which tracker relates to which feature on which War Engine.
 ====
 
-The main difference between the two technologies is that on collapse, power fields are destroyed and of no further use in the battle, while void shields may recover during the battle to provide protection again.
-You check to see if void shields recover at the appropriate step in the Rally phase.
+The main difference between the two technologies is that on collapse, power fields are destroyed and of no further use in the battle, while void shields may be repaired during the battle to provide protection again.
+You check to see if void shields are repaired at the appropriate step in the Rally phase.
+// IJW editing note: I've changed all 'recover' references to 'repair' for consistency with other pages.
```
digi-brain, line 28: Yes, that's good, I'll just remove the editing note.
```suggestion
```

modules/special-rules/pages/war-engine-shields.adoc
```diff
@@ -23,11 +23,13 @@ When all of the unit's fields/shields are collapsed then any further hits will c
...
 
 Neither power fields nor void shields offer any protection from close combat hits, as the enemy is inside the protective envelope at that point.
 {blank}[TODO: " -- though they do protect in a firefight" perhaps?]
+// IJW reply: I don't think the firefight text is needed, as this is effectively a duplicate of the CC page text, and there is no equivalent text on the firefight page.
```
digi-brain, line 32: Okay, it was just a thought to mitigate against the possibility that some people might misconstrue from the text here in the broader association of assaults — but this might be overthinking it. I'm happy to leave it. I'll just queue editing notes for removal.
```suggestion
```

modules/main-rules/pages/roll-for-close-combat-hits.adoc
digi-brain: Oops! Yes.
```suggestion
Don't forget that you can place a maximum of two units in base contact with an enemy unit.
You can choose which enemy units to apply hits against -- but, similar to the allocation of shooting hits, you may not allocate a second hit to a unit unless you have already allocated a hit to every unit that is in base contact.
```

modules/special-rules/pages/shooting-at-war-engines.adoc
digi-brain: Yes, but that text is in the topic that actually does form part of 'this procedure'. The edits we're making here aren't actually in a topic that is 'in this procedure'.

modules/special-rules/pages/shooting-at-war-engines.adoc
```diff
@@ -9,8 +9,8 @@ These categories reflect the fact that these units are so large and easy to hit.
 
 * A War Engine unit can become Immobilised -- and so even easier to hit -- as a result of *critical damage*.
 * There is no target category for 'Immobilised War Engine in the open' -- instead, you simply roll as many hit dice as the total Firepower that you direct at the unit.
-
-{blank}[TODO: Cross-reference the Firepower table, once we've added it. Also, we actually target detachments not units, and we need to clarify what happens when a War Engine detachment has a mix of Immobilised and non-Immobilised WE units.]
+* If a target detachment contains both mobile and Immobilised War Engine units, then you may choose to shoot at only those that are Immobilised, if you wish.
+If you do this then you may allocate hits to the Immobilised units **only**.
```
digi-brain, line 13: Damn - in previous suggestion I accidentally inserted the GitHub double-asterisk markup for bold, rather than the Asciidoc single asterisk markup.

```suggestion
If you do this then you may allocate hits to the Immobilised units *only*.
```

modules/special-rules/pages/war-engine-catastrophic-damage.adoc
IJW-Wartrader: Yeah, I'm not that happy with 'salvage' either.

digi-brain: A couple of ideas for this. I favour the first one, but I could live with the second one if you feel opposed to it.

### (1) Consolidate on the term 'wrecked'

We could simply use the term 'wrecked' in all cases. I don't see why this shouldn't apply to Tyranid bio-constructs too. On their actual data sheets, we can still cast the fluff in terms of 'death' — and this need not conflict with the result of 'Wrecked', in my opinion:

>**1–2 Terminal Bleeding — Wrecked**: The Dominatrix gushes ichor and bodily fluids as it dies in a sodden heap. 

Seems fine to me. And after all, it's no less odd to speak of 'repairs' for bio-constructs — but we don't let that worry us.

I think that this is the most straightforward solution. We could also drop the 'salvageable' bit from the text here, thus:

```suggestion
A destroyed war engine may be only wrecked.
Or, it may be completely annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
```

---

### (2) Just drop the reference to 'salvageable'

```suggestion
A destroyed war engine might be only wrecked (or killed, in the case of the giant Tyranid bio-constructs).
Or, it may be annihilated in a spectacular fashion that presents a hazard to nearby units, whether friend or foe.
```
This seems okay, it's just a bit redundant to have both 'wrecked' and 'killed' as synonyms where the rules are concerned (and of course 'killed' can be rather more general like 'destroyed' can, although I've been careful to avoid that usage).

Not a big problem because, as far as I can recall, we don't really need to refer to the 'Wrecked' or 'Killed' states outside of the data sheets themselves. So we shouldn't need untidy constructions like 'Wrecked or Killed'. The real problem was the multiple different meanings for 'destroyed' when we do actually refer specifically to the 'Destroyed' outcome in the army morale rules — which we solve by use of 'Annihilated'.

modules/special-rules/pages/weapon-mount-examples.adoc
```diff
@@ -25,7 +25,7 @@ For each mount, you can choose a weapon system from a variety of options, as we
 The Imperial Warhound Titan has two weapon mounts:
 
 * *Left Arm* and *Right Arm* (both `Front 90°`).
-Each mount holds a weapon system named Weapon Battery (Firepower `6`).
+Each mount usually holds a weapon system named Weapon Battery (Firepower `6`).
```
digi-brain, line 28:
```suggestion
In this case, each mount holds a weapon system named Weapon Battery (Firepower `6`).
```

modules/special-rules/pages/war-engines-in-firefights.adoc
```diff
@@ -2,5 +2,6 @@
 
 War engines participate in firefights in much the same way as other units.
 
-* Count the Firepower of each weapon system that can draw a line of fire to at least one enemy unit within 15 cm.
+* Count the Firepower of each weapon system that has at least one enemy unit within fire arc and within 15 cm.
+//IJW editing note: firefights aren't restricted by line of fire, so this has to refer to the weapon arc instead.
```
digi-brain, line 6: Ah yes, good catch. Just a tweak to correct 'fire arc' to 'weapon arc' to be consistent with our definition.
```suggestion
* Count the Firepower of each weapon system that has at least one enemy unit within its weapon arc and within 15 cm.
```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
 ====
 {blank}[TODO: Need a suitable graphic.]
 
-In the Assault phase, you make a charge move with your Great Gargant, to assault the Space Marines.
+In the Assault phase, your Great Gargant is in an assault with the Space Marines.
 Your Great Gargant is undamaged, so its current Damage Capacity value is 18.
 This means that in the subsequent close combat, you roll nine dice to hit the Space Marines.
 Your opponent may still only roll a number of dice equal to the number of Space Marine units that are in base contact -- four dice, in this case.
 ====
+// IJW editing note: I've removed the reference to the Gargant making a charge move, because the original diagram showed a situation that required war engines being able to barge enemy units out of the way - and that's currently just an experimental rule!
```
digi-brain, line 58: Makes sense — although of course we could draw our graphic in a different way so that this isn't an issue. But it's still worth implementing the thrust of your idea here I think — just with some further tweaks for precision and consistency throughout the example.

```suggestion
Your Great Gargant is in close combat with some Space Marine infantry.
It is undamaged, so its current Damage Capacity value is `18`.
This means that you roll nine dice to hit the Space Marine units.

Your opponent may still only roll a number of dice equal to the number of Space Marine units that are in base contact -- four dice, in this case.
====
```

modules/special-rules/pages/war-engines-and-blast-markers.adoc
```diff
@@ -1,11 +1,29 @@
 = War engines and Blast markers
 
-Blast markers affect War Engine detachments in much the same way as other units. 
+Blast markers affect War Engine detachments in much the same way as other units.
 However, *all* War Engine detachments have the {stubborn} ability.
 
-Note that for a War Engine detachment has a total initial Damage Capacity value of more than `15`, you may add +1 to the result of the Leadership test.
+You may add +1 to the result of Leadership tests when the total initial Damage Capacity value of the War Engine detachment is `15` or more, instead when it has `15` or more units.
```
digi-brain, line 6: Rationalising on '15 or more' is good, but the phrasing is tricky for a couple of reasons. One reason is that we don't want to imply that a detachment gets the bonus based on the total initial DC as it was at the start of the game, but rather on the total initial DCs of all units that remain in play at this moment.

I think that this should work:

```suggestion
Where your War Engine detachment currently has units in it whose *initial* Damage Capacity values total 15 or more, you may add +1 to the result of the Leadership test.
```

modules/special-rules/pages/war-engines-and-blast-markers.adoc
```diff
@@ -1,11 +1,29 @@
...
+
 {blank}[TODO: Is this the best place to highlight this point? I think it might be better to improve the Leadership test text and to clarify what is what when it comes to 'number of units' and 'detachment strength' and initial Damage Capacity versus current Damage Capacity for the purposes of the Leadership test.]
 
+.Ork Great Gargant
```
digi-brain, line 10: I'm still not sure about this. It'd be better on the one hand to refine the Leadership test topic and eliminate the need to mention it here — but on the other hand it would mean that we'd necessarily have to mention some special rules concepts, which we try not to do in the core rules.

@IJW-Wartrader Do you have any more thoughts on the matter, either way? If not then we could leave this note as-is for now and experiment with the idea another time in future — perhaps when we're getting nearer to a first stable release and we go back over the rules to clear up all TODO notes in one way for another.

modules/special-rules/pages/war-engines-and-blast-markers.adoc
```diff
@@ -1,11 +1,29 @@
 = War engines and Blast markers
 
-Blast markers affect War Engine detachments in much the same way as other units. 
+Blast markers affect War Engine detachments in much the same way as other units.
 However, *all* War Engine detachments have the {stubborn} ability.
 
-Note that for a War Engine detachment has a total initial Damage Capacity value of more than `15`, you may add +1 to the result of the Leadership test.
+You may add +1 to the result of Leadership tests when the total initial Damage Capacity value of the War Engine detachment is `15` or more, instead when it has `15` or more units.
+
 {blank}[TODO: Is this the best place to highlight this point? I think it might be better to improve the Leadership test text and to clarify what is what when it comes to 'number of units' and 'detachment strength' and initial Damage Capacity versus current Damage Capacity for the purposes of the Leadership test.]
```
digi-brain, line 8: I'm still not sure about this. It'd be better on the one hand to refine the Leadership test topic and eliminate the need to mention it here — but on the other hand it would mean that we'd necessarily have to mention some special rules concepts, which we try not to do in the core rules.

@IJW-Wartrader Do you have any more thoughts on the matter, either way? If not then we could leave this note as-is for now and experiment with the idea another time in future — perhaps when we're getting nearer to a first stable release and we go back over the rules to clear up all TODO notes in one way for another.

digi-brain: Actually, given your addition of examples below, I think it makes most sense to do it all as we've done here — otherwise we kinda need to add the examples to the core rules as well. So, might as well delete this note I think.
```suggestion
```

modules/special-rules/pages/war-engines-and-blast-markers.adoc
```diff
@@ -1,11 +1,29 @@
...
+.Imperial Guard Super Heavy Tank Company
+====
+Baneblades are war engines with an initial Damage Capacity value of 4.
+
+A detachment of four Baneblades has a total initial Damage Capacity value of 16, so you may add +1 to the result of Leadership tests.
+
+If the detachment is reduced to three Baneblades the total is 12, so you cannot add +1 to the result of Leadership tests.
+====
+
+
```
digi-brain, line 26: Please double-check that the tweaks in this suggestion don't 'undo' anything or introduce new problems.

_Aside: Note that I use the backtick 'code' format for actual data sheet values, and not for totals of multiple data sheet values._

```suggestion
.Leadership test bonus -- Ork Great Gargant
====
The Great Gargant is a War Engine unit that has an initial Damage Capacity value of `18`.
It forms its own detachment of one unit.

You may add +1 to the results of Leadership tests for this detachment, even after the Great Gargant's current Damage Capacity value falls below 15.
====

.Leadership test bonus -- Imperial Guard Super Heavy Tank Company
====
The Baneblade is a War Engine unit that has an initial Damage Capacity value of `4`.
So, a detachment that currently includes four Baneblade units has an initial Damage Capacity value total of 16.

You may add +1 to the results of Leadership tests for this detachment, even after their current Damage Capacity values total less than 15.

But, if any units are destroyed or no longer in play for some reason, then the initial Damage Capacity values will total less than 15.
So then you may no longer add +1 to the results of Leadership tests.
====

```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
-You can 'gang up' on War Engine units with a number of regular units equal to half of the War Engine unit's initial Damage Capacity value.
-{blank}[TODO: How do War Engine units -- or a mix of War Engine and regular units -- gang up on on other War Engine units?]
+You can place as many units in base contact with the War Engine unit as its initial Damage Capacity value.
+
+[NOTE.e40k]
+====
+In _Epic 40,000_, you could only 'gang up' with half as many units, which meant that smaller War Engine units like Baneblades could only be attacked by two units -- the same as a regular target!
+====
+//{blank}[TODO: How do War Engine units -- or a mix of War Engine and regular units -- gang up on on other War Engine units?]
+// IJW reply just treat them as units, so that you can actually gang up on similar-size targets.
```
digi-brain, line 18: Yup, that makes sense -- I had some thought in mind while I was editing here, about how to apply the DCs on each side, but that was erroneous. Queuing for removal.
```suggestion
```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
 Lesser units can and must 'gang up' on a War Engine if they are to defeat it in close combat.
 
-You can 'gang up' on War Engine units with a number of regular units equal to half of the War Engine unit's initial Damage Capacity value.
-{blank}[TODO: How do War Engine units -- or a mix of War Engine and regular units -- gang up on on other War Engine units?]
+You can place as many units in base contact with the War Engine unit as its initial Damage Capacity value.
+
+[NOTE.e40k]
+====
+In _Epic 40,000_, you could only 'gang up' with half as many units, which meant that smaller War Engine units like Baneblades could only be attacked by two units -- the same as a regular target!
+====
```
digi-brain, line 16: Given that strictly speaking this weakens war engines to some extent, is there any other reason for this rule change? 

I ask because I imagine that others may not see "we can't 'gang up' any more on small war engines than we can on regular units" as a problem — or not enough of a problem to justify doubling the number of units that can be in contact for all war engines.

Meantime, just a few minor tweaks:

```suggestion
[NOTE.e40k]
====
In _Epic 40,000_, you could only 'gang up' with half as many units as the initial Damage Capacity.
This meant that against small War Engine units such as the Baneblade you could only place two units in base contact -- the same as against a Vehicle or Infantry target!
====
```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
-You may move your War Engine units in the Assault phase, and you may make charge moves with them if you wish. 
+You may move your War Engine units in the Assault phase, and you may make charge moves with them if you wish.
 This does not affect the War Engine unit's ability to shoot in the Shooting phase in any way; it always fires at full effect.
 
 War Engine units have a comparatively high Assault value, as you will see on their data sheets.
 Lesser units can and must 'gang up' on a War Engine if they are to defeat it in close combat.
 
-You can 'gang up' on War Engine units with a number of regular units equal to half of the War Engine unit's initial Damage Capacity value.
-{blank}[TODO: How do War Engine units -- or a mix of War Engine and regular units -- gang up on on other War Engine units?]
+You can place as many units in base contact with the War Engine unit as its initial Damage Capacity value.
```
digi-brain, line 11:
```suggestion
The maximum number of units that you may place in base contact with a War Engine unit is equal to the number of its initial Damage Capacity value.
```
...though see my next comment about the actual rule change.

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
+// IJW reply just treat them as units, so that you can actually gang up on similar-size targets.
 
 == Close combat attacks with a War Engine
 
 War engines can also do a lot more damage in close combat than regular units.
 
 * Roll a number of dice to hit equal to half the current Damage Capacity of the War Engine unit (round up).
-* Note that you may allocate hits only among units that the War Engine is actually in base contact with.
-So, even if you score excess hits, your War Engine unit can't destroy more units in close combat than it is actually in base contact with.
+* As normal, you can apply hits to any enemy unit in the combat which is in base contact with any of your units.
```
digi-brain, line 25:
```suggestion
* As normal, you can allocate hits to any enemy units that are in base contact with any of your units in the close combat.
```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
 
-{blank}[TODO: In a detachment of multiple War Engine units, can units that are not in base contact with the enemy provide supporting fire like regular units can? If so, is it just +1 to the Assault value total, or should it be '+half the current Damage Capacity' as it is when they support *other* detachments?]
+[NOTE.e40k]
+====
+In _Epic 40,000_, the rules for close combat attacks with War Engines assumed that you had a single War Engine in combat with the enemy, and that there were no other War Engines or allied detachments in the close combat. This revised version treats all the close combat hits normally, instead of limiting the hits to units in base contact with that specific War Engine.
+====
```
digi-brain, line 30: Some tricky phrasing again — please check that I don't 'break' anything with this suggestion.
```suggestion
[NOTE.e40k]
====
In _Epic 40,000_, the rules for close combat attacks with War Engines are worded on the basis of a single War Engine in close combat with the enemy, with no other help.
In the absence of further clarification, this tends to imply that the allocation of hits might be on a per War Engine basis, in contrast to the usual rules.
We word the rules to make it clear that we should allocate all close combat hits in the same way, regardless of origin.
====
```

modules/special-rules/pages/war-engines-in-close-combat.adoc
```diff
@@ -2,40 +2,61 @@
...
+
+* When a War Engine detachment is in close combat and some of the War Engine units are not in base contact with the enemy, do not add +1 for the units not in base contact.
+Instead, add a number equal to half the current Damage Capacity of these War Engines.
+This does not prevent the War Engine units from supporting other close combats.
+
+[NOTE.e40k]
+====
+In _Epic 40,000_, there were rules for War Engines to support *other* close combats, but no rules for supporting a close combat that the War Engine detachment was in. We have adapted the rules for supporting other close combats.
+====
+
```
digi-brain, line 48: Some tweaks — please check that I don't 'break' anything.
```suggestion
== War Engine detachments in close combat

* When you add up Assault values for a War Engine detachment that is in close combat, do not add +1 for 'supporting fire' from each unit in the detachment that is within 15 cm of the enemy but not in base contact.
Instead, add a number equal to half the current Damage Capacity of each such War Engine unit.
+
To be clear, such units may still also support *other detachments* as we explain at xref:war-engines-in-support-of-close-combats.adoc[].

[NOTE.e40k]
====
In _Epic 40,000_, the rules specify how War Engine units can support other detachments that are in close combat, but not how they can provide supporting fire within their own detachment.
We make both situations plain.
====

```

modules/special-rules/pages/war-engines-in-close-combat.adoc
IJW-Wartrader: The previous discussion appeared to be strongly in favour of increasing it: https://www.facebook.com/groups/1004227353047038/posts/2267032236766537/

But I've started the usual polls to see if people disagree...

digi-brain: Some further tweaks: You make a good case on Facebook, and I agree — but I think that the particular point you used in this note was one of the weakest ones. While we don't want to go into depth here, I think that it'd be best to allude to the stronger reasons.

Something like this perhaps:

```suggestion
[NOTE.e40k]
.'Ganging up' on War Engine units
====
In _Epic 40,000_, you can only 'gang up' with *half* as many units as the initial Damage Capacity.
This leads to a number of odd effects, not least that it makes war engines *excessively* difficult to overcome in close combat -- so, we reference the full initial Damage Capacity instead.
====
```

modules/special-rules/pages/war-engine-catastrophic-damage.adoc
digi-brain: I think I'll go ahead and commit my suggestion (1). After all, special variation of 'Wrecked' terminology in favour of 'Killed' seems doubly needless when we'll still call the Tyranid bio-constructs 'war engines' despite the fact that they aren't really mechanical 'engines' like the rest...

modules/special-rules/pages/war-engines-in-close-combat.adoc
digi-brain: This appears to contradict other statements and is thus 'wrong': `This does not prevent the War Engine units from supporting other close combats.`

The original rules state `In order to lend its support to a close combat a war engine must be within 15cm of an enemy unit which is fighting in, or lending support to, the combat and the war engine can't be in close combat itself.` ...and our current Remastered rules mirror this with `To do this, your War Engine unit must not be engaged in combat itself`.

So, I'll tweak my suggestion here a bit further, thus: (along with some other further small tweaks while I'm at it)

```suggestion
== War Engine detachments in close combat

When you add up Assault values for a War Engine detachment that is in close combat, do not add +1 for 'supporting fire' from each unit in the detachment that is within 15 cm of the enemy but not in base contact.
Instead, add a number equal to half the current Damage Capacity of each such War Engine unit.

[NOTE.e40k]
.Supporting fire from War Engine units 
====
_Epic 40,000_ gives rules for War Engine units in support of other detachments that are in close combat -- but does not gives rules for how they can provide supporting fire within their own detachment.
We make both situations plain.
====

```

IJW-Wartrader: > This appears to contradict other statements and is thus 'wrong': `This does not prevent the War Engine units from supporting other close combats.`

Could you explain in a bit more detail? I can't see where this is contradicting other statements. As a reader, I no longer know how this affects War Engines supporting other close combats.

digi-brain: The detail I already added above covers it, doesn't it?

> The original rules state `In order to lend its support to a close combat a war engine must be within 15cm of an enemy unit which is fighting in, or lending support to, the combat and the war engine can't be in close combat itself.` ...and our current Remastered rules mirror this with `To do this, your War Engine unit must not be engaged in combat itself`.

...where original is [War engines supporting close combats](https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/WarEngines/WarEnginesSupportingCloseCombats.htm) and Remastered is [War engines in support of close combats](https://epicremastered.com/core/special-rules/war-engines-in-support-of-close-combats.html).

Note "can't be in close combat itself" and "must not be engaged in combat itself", respectively. Unless I'm missing something daft (always possible!) these seem to directly contradict that particular line that I've excluded from my suggestion.

IJW-Wartrader: I think the missing bit is that close combats are resolved sequentially rather than simultaneously.

So a War Engine detachment can be in a close combat (with one of them outside base contact), resolve their own close combat, and then be in a position to support another separate close combat that hasn't been resolved yet. 

Without the removed bullet, it's unclear whether the 'supporting' War Engine can support its own detachment and then support the separate close combat.

digi-brain: Ah, okay. So, resolve all close combats, but choose the order carefully to maximise support from war engines... don't treat close combats as simultaneous.

I made an implicit assumption that although of course we resolve them one at a time, we nonetheless treat them as simultaneous. As we are **not** supposed to do that, **should we make this more clear somewhere**? (Others might make the same assumption that I did.) Or is it already clear somewhere and I've just overlooked/forgotten statements to that effect?

Latest iteration of the suggestion:

```suggestion
== War Engine detachments in close combat

When you add up Assault values for a War Engine detachment that is in close combat, do not add +1 for 'supporting fire' from each unit in the detachment that is within 15 cm of the enemy but not in base contact.
Instead, add a number equal to half the current Damage Capacity of each such War Engine unit.

To be clear, such units may still also support *other detachments* as we explain at xref:war-engines-in-support-of-close-combats.adoc[].

[NOTE.e40k]
.Supporting fire from War Engine units 
====
_Epic 40,000_ gives rules for War Engine units in support of other detachments that are in close combat -- but does not give similar rules to apply the same logic for supporting fire within their own detachment (so only the usual +1 would apply, which seems odd).
We make both situations plain.
====
```

digi-brain: @IJW-Wartrader Any thoughts on my last comment on this thread? If not then I'll just apply this suggestion and proceed to final checks with a view to publish, as there are no other outstanding items.

IJW-Wartrader: @digi-brain that looks OK.

> I made an implicit assumption that although of course we resolve them one at a time, we nonetheless treat them as simultaneous. As we are **not** supposed to do that, **should we make this more clear somewhere**? (Others might make the same assumption that I did.) Or is it already clear somewhere and I've just overlooked/forgotten statements to that effect?

Hmm. Not sure. For me, 'The player with the Initiative decides the order in which to resolve close combats.' from close-combat-procedure.adoc makes it clear that you resolve one CC, then move onto the next and resolve that, etc.

digi-brain (approved):
Final checks complete, ready to merge.

modules/special-rules/pages/war-engines-in-close-combat.adoc
IJW-Wartrader: > Well, my point is that **of course** you must 'decide the order' to resolve one at a time, because that is the **only** practical way to do it in a tabletop wargame. So, by itself, this:
> 
> > The player with the Initiative decides the order in which to resolve close combats.
> > ...
> > To resolve a close combat, use the close combat procedure as follows.
> 
> ...doesn't preclude the possibility that we treat the set of combats that a player triggers as simultaneous events.

If you think it's unclear, then sure, add something. But I may not have explained myself properly.

Screen Shot 2022-01-04 at 16 01 54

You choose the order in which to resolve the combats (first red underline). Then you resolve **a combat** (second red underline), following the six steps (green bar). This includes one of the detachments losing and making a Retreat move, physically moving the models. Then you go to the next combat and repeat the six steps. Everything that happened in the first combat has been resolved and can potentially affect the second combat. In particular, where the losers retreated to and how this affects any 'new' Retreat moves as there may no longer be enough space to retreat into. 

I can't see a way to read the current rules text in a different way. For example, Retreat moves would have to be totally rewritten so that you check the 15 cm distance at the end of the phase, instead of the end of the Retreat move.

PR #40, opened 2023-01-18
Fix rally test text

digi-brain:
@IJW-Wartrader I think I've fixed the main problem as discussed, but I've also added some new TODO notes on awkward things that occurred to me.

I also made a few other small tweaks while I'm at it — hopefully these are improvements as I intended and don't cause new problems. 😅

See what you make of it.

### DIFF
--- a/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
+++ b/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
@@ -3,18 +3,20 @@
 If you can rally a Broken detachment then you will regain control so that you can issue orders to it again next turn.
 
 For each Broken detachment that you have, take a Rally test to see if they rally.
-This is a standard Leadership test.
-
-.Rally test
-Make a standard Leadership test -- all the usual rules for modifiers and such apply.
-If you succeed then the detachment rallies -- remove the Broken condition marker from the detachment; otherwise, the detachment remains Broken -- try again in the next Rally phase.
 
 [IMPORTANT]
 ====
-You cannot rally a detachment that has as many Blast markers (or more) as it has units remaining -- you may not attempt the test in such cases.
-
-You may be able to take the test for such detachments at the end of a future turn if you are able to remove enough Blast markers at step xref:roll-to-remove-blast-markers.adoc[].
+You may only attempt a Rally test when a Broken detachment has fewer Blast markers than the number of units it has remaining.
+{blank}[TODO: I think it'd be better to phrase this along the lines of "fewer Blast markers than its current detachment strength", to align with how we've defined this for {robust} units.]
 ====
+
+.Rally test
+Make a standard Leadership test -- all the usual rules for modifiers and such apply.
+
+* If you succeed then the detachment rallies -- remove the Broken condition marker from the detachment;
+* If you fail then the detachment remains Broken -- try again in the next Rally phase. 
+
+Either way, leave the Blast markers in place.
 
 .Related information
 * xref:broken-detachments.adoc[]
--- a/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
+++ b/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
@@ -7,7 +7,6 @@ For each Broken detachment that you have, take a Rally test to see if they rally
 [IMPORTANT]
 ====
 You may only attempt a Rally test when a Broken detachment has fewer Blast markers than the number of units it has remaining.
-{blank}[TODO: I think it'd be better to phrase this along the lines of "fewer Blast markers than its current detachment strength", to align with how we've defined this for {robust} units.]
 ====
 
 .Rally test
--- a/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
+++ b/modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
@@ -3,15 +3,19 @@
 If you can rally a Broken detachment then you will regain control so that you can issue orders to it again next turn.
 
 For each Broken detachment that you have, take a Rally test to see if they rally.
+This is a standard Leadership test.
 
 .Rally test
-Roll a D6: If you score higher than the number of Blast markers on the detachment then you succeed -- remove the Broken condition marker from the detachment; otherwise, you fail -- leave the marker in place.
+Make a standard Leadership test -- all the usual rules for modifiers and such apply.
+If you succeed then the detachment rallies -- remove the Broken condition marker from the detachment; otherwise, the detachment remains Broken -- try again in the next Rally phase.
 
 [IMPORTANT]
 ====
-You cannot rally a detachment that has as many Blast markers (or more) as it has units remaining.
+You cannot rally a detachment that has as many Blast markers (or more) as it has units remaining -- you may not attempt the test in such cases.
 
 You may be able to take the test for such detachments at the end of a future turn if you are able to remove enough Blast markers at step xref:roll-to-remove-blast-markers.adoc[].
 ====
 
-NOTE: The rally test works in a similar way to the Leadership test -- but it is different because there are no modifiers and no exceptions for a natural 6.
+.Related information
+* xref:broken-detachments.adoc[]
+* xref:leadership-test.adoc[]
--- a/modules/main-rules/pages/leadership-test.adoc
+++ b/modules/main-rules/pages/leadership-test.adoc
@@ -5,13 +5,11 @@ When your detachment has one or more Blast markers on it, you must pass a Leader
 {blank}[TODO: What happens if you try to set Overwatch orders and fail? Default to Advance orders I presume, and lose the chance to use any other orders? If you pass the test to *set* Overwatch orders, do you get to *move* up to 5 cm under those orders without a further test?]
 
 NOTE: A detachment that is subject to the Broken condition is not under orders -- so, a Retreat move does not trigger a Leadership test.
-{blank}[TODO: There's at least one special case to consider — the {hit-and-run} ability's use of the Retreat move' Should this ability require a Leadership test if the detachment has Blast markers, like a normal move would? If yes then it'd be best to adjust this note to accomodate it better, or possibly change the {hit-and-run} text a bit.]
 
 .Leadership test
 Roll a D6: If you score higher than the number of Blast markers that are on the detachment then you succeed -- you may make whatever moves the orders permit; otherwise you fail -- you may not make any moves with the detachment during this phase.
 
 * If your detachment currently has 15 or more units in it, add +1 to the score.
-{blank}[TODO: Should we phrase this in terms of detachment strength rather than units, to align with how we've defined things for the {robust} ability?]
 * A roll of a natural 6 is always a success, no matter how many Blast markers are on the detachment.
 
 .Related information
--- a/modules/main-rules/pages/leadership-test.adoc
+++ b/modules/main-rules/pages/leadership-test.adoc
@@ -4,12 +4,13 @@ When your detachment has one or more Blast markers on it, you must pass a Leader
 
 {blank}[TODO: What happens if you try to set Overwatch orders and fail? Default to Advance orders I presume, and lose the chance to use any other orders? If you pass the test to *set* Overwatch orders, do you get to *move* up to 5 cm under those orders without a further test?]
 
-NOTE: A Broken detachment is not under orders and no Leadership test is required to make a Retreat move (see xref:broken-detachments.adoc[]).
+NOTE: A detachment that is subject to the Broken condition is not under orders -- so, no Leadership test is required to make a Retreat move.
 
 .Leadership test
 Roll a D6: If you score higher than the number of Blast markers that are on the detachment then you succeed -- you may make whatever moves the orders permit; otherwise you fail -- you may not make any moves with the detachment during this phase.
 
-* If you detachment currently has 15 or more units in it, add +1 to the score.
+* If your detachment currently has 15 or more units in it, add +1 to the score.
 * A roll of a natural 6 is always a success, no matter how many Blast markers are on the detachment.
 
-{blank}[TODO: Rationalise with the Rally phase -- the 'natural 6 is a pass' bit doesn't apply then, and perhaps the +1 bonus for 15 units or more doesn't either?]
+.Related information
+* xref:broken-detachments.adoc[]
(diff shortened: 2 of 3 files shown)


modules/special-rules/pages/hit-and-run.adoc
```diff
@@ -8,13 +8,13 @@ Some forces have honed tactics in which they make a lightning assault, seemingly
 
 Where your detachment has the {hit-and-run} ability:
 
-* After a close combat or firefight with this detachment, you can always make a retreat move -- even if you win.
+* After a close combat or firefight with this detachment, you can always make a Retreat move -- even if you win.
 +
-+[TODO: As written, it seems that they can make the retreat move even if they were charged by the enemy first. Is this intended? It doesn't seem to fit with the idea of 'hit and run'.]+
++{blank}[TODO: As written, it seems that they can make the Retreat move even if they were charged by the enemy first. Is this intended? It doesn't seem to fit with the idea of 'hit and run', it's more like 'hit, or get hit, and run'...!]+
```
digi-brain, line 13: Note, I'm really just asking the question here, I don't feel all that strongly about it — I can see that the 'who actually charged on this occasion' bit is somewhat arbitrary given the nature of the game mechanics and turn structure.

modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
```diff
@@ -4,14 +4,20 @@ If you can rally a Broken detachment then you will regain control so that you ca
...
-.Rally test
-Roll a D6: If you score higher than the number of Blast markers on the detachment then you succeed -- remove the Broken condition marker from the detachment; otherwise, you fail -- leave the marker in place.
-
 [IMPORTANT]
 ====
-You cannot rally a detachment that has as many Blast markers (or more) as it has units remaining.
-
-You may be able to take the test for such detachments at the end of a future turn if you are able to remove enough Blast markers at step xref:roll-to-remove-blast-markers.adoc[].
+You may only attempt a Rally test when a Broken detachment has fewer Blast markers than the number of units it has remaining.
+{blank}[TODO: I think it'd be better to phrase this along the lines of "fewer Blast markers than its current detachment strength", to align with how we've defined this for {robust} units.]
```
digi-brain, line 10: This is one of two instances like this that I've spotted — there may be more. On the one hand it'd be simpler and more intuitive to stick with units — but on the other hand, given the effect that _Robust_ has on detachment strength, I figure it'd be more consistent to base it on detachment strength so we don't have a disparity between half-strength for army morale and number of units for Rally tests...

IJW-Wartrader: Agreed, it should be detachment strength. But...

As far as I can see, 'detachment strength' is currently an undefined term and only appears on the Robust page. I assume this will be expanded on in the battles section for Army Morale?

modules/special-rules/pages/hit-and-run.adoc
IJW-Wartrader: My reply must have got lost in an old pull request somewhere!

The original text just specifies 'after a close combar or firefight, even if the win', so there's no requirement for the detachment to be the one assaulting. I don't see any background issues as it still represents the detachment fading away from an assault.

modules/main-rules/pages/leadership-test.adoc
```diff
@@ -4,12 +4,15 @@ When your detachment has one or more Blast markers on it, you must pass a Leader
 
 {blank}[TODO: What happens if you try to set Overwatch orders and fail? Default to Advance orders I presume, and lose the chance to use any other orders? If you pass the test to *set* Overwatch orders, do you get to *move* up to 5 cm under those orders without a further test?]
 
-NOTE: A Broken detachment is not under orders and no Leadership test is required to make a Retreat move (see xref:broken-detachments.adoc[]).
+NOTE: A detachment that is subject to the Broken condition is not under orders -- so, a Retreat move does not trigger a Leadership test.
+{blank}[TODO: There's at least one special case to consider — the {hit-and-run} ability's use of the Retreat move' Should this ability require a Leadership test if the detachment has Blast markers, like a normal move would? If yes then it'd be best to adjust this note to accomodate it better, or possibly change the {hit-and-run} text a bit.]
```
IJW-Wartrader, line 8: 'can always' in the original text, for me, infers no requirement for a Leadership test.

And it would also cause extra problems - what happens if they choose to make a retreat move and fail the Leadership test? They've chosen to make a retreat move, so everyone within 15 cm of an enemy would die. :-(

modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
digi-brain: **TL;DR**: We could skip over this query for now and just keep the unit-based text for the time being.

**Long version**:

Yes, to date we've only defined detachment strength as part of the the _Robust_ rules. I was thinking ahead a bit here, aiming to future-proof the text somewhat.

So, to think it through some more:

When we/I get around to adding drafts of the Battles book/Armies book content then 'detachment strength/half-strength' will necessarily feature as part of the explanation of the Army Morale rules/detachment card explanation. This concept already exists in the original rules, just in content we haven't gotten to yet because we've only completed a pass through the Core component, not Battles and Armies. (And of course we'll give it a more distinct definition.)

The experimental _Robust_ ability rules text explains how this ability should integrate with that concept — because this is a new idea that didn't exist in the source and so wasn't accounted for.

The IMPORTANT note for the Rally test a question of how to apply _Robust_ in this context — and it seems like the the obvious and most consistent solution is to unify with the detachment strength concept and use the same terminology.

But, given that the _Robust_ ability is still experimental at this point in time, perhaps the best path for the moment is to:

1. Stick with the unit-based phrasing here for the moment.
2. Tweak/expand the experimental _Robust_ rules text to make the Rally test situation plain as well. Or at least add a TODO note there so that we have a reminder and players have a pointer. (Either way, we should probably create an issue so we don't forget, and then a branch to implement it.)
3. Revisit if/when we add the other content and/or make _Robust_ 'official' in future, and integrate as appropriate.

What do you reckon?

modules/main-rules/pages/leadership-test.adoc
digi-brain: That seems pretty clear now, I guess I was over-thinking it in the moment of editing.

Good point — not only would a Leadership test for the _Hit and Run_ case clash with the NOTE here and require a special exception, there's the problem you've described.

I'm happy to declare my TODO note totally redundant in this case and resolve this thread. 😅

```suggestion
```

modules/main-rules/pages/leadership-test.adoc
```diff
@@ -4,12 +4,15 @@ When your detachment has one or more Blast markers on it, you must pass a Leader
...
 
 .Leadership test
 Roll a D6: If you score higher than the number of Blast markers that are on the detachment then you succeed -- you may make whatever moves the orders permit; otherwise you fail -- you may not make any moves with the detachment during this phase.
 
-* If you detachment currently has 15 or more units in it, add +1 to the score.
+* If your detachment currently has 15 or more units in it, add +1 to the score.
+{blank}[TODO: Should we phrase this in terms of detachment strength rather than units, to align with how we've defined things for the {robust} ability?]
```
digi-brain, line 14: Pretty much the same applies here as at https://github.com/GameBrains/er-core/pull/40/files#r1083451747, I think.

modules/special-rules/pages/hit-and-run.adoc
digi-brain: Thematically, I do think that there is a difference between being the ones to initiate an assault and then fading away, and being assaulted and then fading away. Mechanically, we can draw a parallel in which side actually makes the assault/charge move — which would make sense as the 'Hit' part of _Hit and Run_.

So, I wonder if the ability should trigger only when the controlling player made the assault/charge move into firefight/close combat.

The problem is that the [original text](https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/SpecialistsAndSpecialAbilities/HitAndRun.htm) is somewhat ambiguous, because the main paragraph uses thematic phrasing and casts White Scars as the prime movers throughout — and it doesn't actually describe things in terms of game mechanics and player actions:

> The White Scars attack from nowhere, unleash a devastating volley of fire or brief charge into close combat, and then disappear as suddenly as they arrived. To represent this, White Scars detachments can always make a retreat move after a close combat or firefight, even if they win. This enables them to attack the enemy and then withdraw to safety to avoid an unfavourable assault or firefight from other enemy detachments close by.

To my eyes, you can easily interpret thus:

- "attack from nowhere" — they initiate the assault/the _Hit and Run_ controlling player makes the assault move
- "brief charge" — they initiate the close combat/the _Hit and Run_ controlling player makes the charge move
- "they arrived" — they made the assault/the _Hit and Run_ controlling player made the assault move
- "enables them to attack the enemy and then withdraw" — they initiate the assault and then get to withdraw/the _Hit and Run_ controlling players makes the assault move and may make a Retreat move even if they win (which we might take to imply that the latter is dependent upon the former)

If you say that this interpretation is too restrictive or otherwise problematic in the game mechanics or game balance, and that _Hit and Run_ controlling players should be able to make a Retreat move after they win whether nor not they made the assault move to initiate that assault, that's fine with me — I'm just asking the question because it seems ambiguous enough to allow argument for either interpretation.

At the extreme, think of this possibility if the _Hit and Run_ controlling player can have it both ways:

1. The _Hit and Run_ detachment is in a static, defensible position. (Not what you'd normally expect, but possible.)
2. The opponent assaults this detachment, moving their forces into firefight/close combat.
3. The _Hit and Run_ player wins the assault, repelling the enemy forces — and then uses the ability to leave that static position.

This seems odd to me. We can explain it away if we have to, but given the ambiguity of the original text and the thematic oddity of this situation, I figure it bears some consideration.

modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
IJW-Wartrader: I'd be happy with 2. It was more a comment that we haven't defined detachment strength yet, than that something needed changing.

modules/main-rules/pages/leadership-test.adoc
IJW-Wartrader: Sounds good to me.

modules/special-rules/pages/hit-and-run.adoc
IJW-Wartrader: > The problem is that the [original text](https://thehobby.zone/resources/e40k-compendium/Content/SpecialRules/SpecialistsAndSpecialAbilities/HitAndRun.htm) is somewhat ambiguous, because the main paragraph uses thematic phrasing and casts White Scars as the prime movers throughout — and it doesn't actually describe things in terms of game mechanics and player actions:

Ah! I was going by the bit that **is** written in terms of game mechanics, and ignoring everything that (to me) looks like background or description (mechanics marked in bold):

>> _The White Scars attack from nowhere, unleash a devastating volley of fire or brief charge into close combat, and then disappear as suddenly as they arrived. To represent this,_ **White Scars detachments can always make a retreat move after a close combat or firefight, even if they win**. _This enables them to attack the enemy and then withdraw to safety to avoid an unfavourable assault or firefight from other enemy detachments close by._

This gives us:

(Hit and Run) detachments can always make a retreat move after a close combat or firefight, even if they win.

> If you say that this interpretation is too restrictive or otherwise problematic in the game mechanics or game balance, and that _Hit and Run_ controlling players should be able to make a Retreat move after they win whether nor not they made the assault move to initiate that assault, that's fine with me — I'm just asking the question because it seems ambiguous enough to allow argument for either interpretation.

For me, the bit that's actually written in terms of game mechanics isn't ambiguous.

It's simple, it works, I don't think we add anything to the game by making the ability more complex or more restricted.

> At the extreme, think of this possibility if the _Hit and Run_ controlling player can have it both ways:
> 
> 1. The _Hit and Run_ detachment is in a static, defensible position. (Not what you'd normally expect, but possible.)
> 2. The opponent assaults this detachment, moving their forces into firefight/close combat.
> 3. The _Hit and Run_ player wins the assault, repelling the enemy forces — and then uses the ability to leave that static position.
> 
> This seems odd to me. We can explain it away if we have to, but given the ambiguity of the original text and the thematic oddity of this situation, I figure it bears some consideration.

It's possible, as you say, but it's very unlikely as the ability is only on detachments that are fully mechanised and/or made up of very fast units.

But even then, that doesn't sound like a thematic issue. White Scars temporarily taking up a defensive position to lure the attacker into a bad assault, and then feinting away again after the assault sounds completely in character.

modules/main-rules/pages/attempt-to-rally-broken-detachments.adoc
digi-brain: Just to be clear, I meant that as a list of three actions to take in combination, not three separate options. I've now created issue #41 to this effect so we don't lose sight of it, and thus we can clean up **this** PR a bit:

```suggestion
```

modules/main-rules/pages/leadership-test.adoc
digi-brain:
```suggestion
```

modules/special-rules/pages/hit-and-run.adoc
digi-brain: > Ah! I was going by the bit that is written in terms of game mechanics, and ignoring everything that (to me) looks like background or description (mechanics marked in bold):
> [...]
> For me, the bit that's actually written in terms of game mechanics isn't ambiguous.

My point is that it wouldn't be the first time that GW text failed to mention an important detail in the actual explicit rules bit that we had to infer from the surrounding descriptive bit — and that they actually **intended** that interpretation. Thus, ambiguous.

I get that we can't spend all our time hypothesising about things that GW could've failed to mention — but I raised the question here specifically because because all the surrounding descriptive text is very much in the sense that the _Hit and Run_ detachment initiated the assault.

> It's simple, it works, I don't think we add anything to the game by making the ability more complex or more restricted.

I don't think that restriction would complicate matters any more than many other simple restrictions that exist elsewhere in the game. **But**, if you don't think this is an occasion where GW made a mistake of the type I mention above, and you don't think there's any benefit to such a restriction anyway, then that's fine with me. In that case we can delete this TODO note.

```suggestion
```

digi-brain:
Completed final checks and tweaks — ready to publish.

PR #43, opened 2023-04-24
Pulsar x first pull

IJW-Wartrader:


modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -1,18 +1,19 @@
...
+The weapon shoots in rapid but powerful bursts, guided by advanced targeting technologies.
 As a result, the weapon can destroy multiple targets in a single salvo.
 
 ---
 
 When you shoot with a {pulsar}:
 
-* Roll a D6 and resolve this many {macro-weapon} shots against the target detachment.
+. Roll the dice listed in brackets, for example (D3+1) or (D6).
+. Resolve that many {macro-weapon} shots against the target detachment.
```
digi-brain, line 13: 'Dice' is plural, and thus far we expect only a single, hence it should be 'die'. However, for a couple of small reasons, I think it'd be better to rephrase without reference to die/dice anyway. So, how about this:

```suggestion
. Determine the number of shots, based on the value that we state in brackets — for example, `(D3+1)` or `(D6)`.
. Resolve that many {macro-weapon} shots against the target detachment.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
...
+As a result, the weapon can destroy multiple targets in a single salvo.
+
+---
+
+When you shoot with a {pulsar-x}:
+
+. Roll the dice listed in brackets, for example (D3+1) or (D6).
+. Resolve that many {macro-weapon} shots against the target detachment.
+
+IMPORTANT: Remember that a {pulsar-x} is a single super-heavy weapon.
```
digi-brain, line 15: To avoid the weirdness of "a _Pulsar (X)_" (and to reinforce the main point slightly while I'm at it), perhaps something like this

```suggestion
IMPORTANT: Remember that {pulsar-x} always represents a single super-heavy weapon, regardless of the number of shots.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
...
+It does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, or contribution to firefights.
```
digi-brain, line 16: Just a thought, and I'm not in any way wedded to this, but might a more concise version be slightly better?

```suggestion
It does *not* count as multiple super-heavy weapons for any game purpose.
```

or perhaps:

```suggestion
It does *not* count as multiple super-heavy weapons for any game purpose, such as placement of Blast markers.
```

If you think that the explicit list you already drafted is better for any reason — and you're confident that there aren't any other list items that we're forgetting — then we can leave this sentence as it is.

modules/special-rules/pages/pulsar-x.adoc
IJW-Wartrader: As much as I would love to use 'die/dice', there are enough dictionaries that now list 'dice' as both singular and plural that I think we've lost that fight, as GW have been using 'dice' for both since the Eighties. 🙁

But as you say, it'd be better not to use it anyway.

'the value we state' reads a bit oddly, given that the value will be read from a data sheet. Maybe this?

```suggestion
. Determine the number of shots, based on the value on the unit's data sheet — for example, `(D3+1)` or `(D6)`.
. Resolve that many {macro-weapon} shots against the target detachment.
```

modules/special-rules/pages/pulsar-x.adoc
IJW-Wartrader: Ooh, That's a bit of a tough one. We wouldn't say 'Remember that Death Ray always represents...', so I think we should use 'Remember that a {pulsar}...'.

So this?

```suggestion
IMPORTANT: Remember that a {pulsar} always represents a single super-heavy weapon, regardless of the number of shots.
```

IJW-Wartrader: P.S. I did a minimal edit as the existing text had already been checked.

modules/special-rules/pages/pulsar-x.adoc
IJW-Wartrader: I don't remember who drafted that bit originally, but I'd prefer to keep it as an explicit list. Mainly because it gets forgotten in firefights.

If anything, I'm inclined to make it more comprehensive by combining the old and new versions:

```suggestion
It does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

modules/special-rules/pages/pulsar-x.adoc
digi-brain: I take your point that the incorrect usage is becoming normalised. Here though, we have the luxury that we have full control over the project, and I prefer to take advantage of that to do it right 😁 Indeed, GW has been getting this consistently wrong for a long time — and it still bugs me every time I see it! 😂

> 'the value we state' reads a bit oddly, given that the value will be read from a data sheet.

In this context 'state' clearly means 'in print', not 'out loud' — that's not uncommon in my experience, and seems perfectly okay to me. (Similarly, a 'statement' is often printed content that people read, not something that they hear someone actually vocalise.)

We do also use 'that we state' in similar fashion in the _Transport (X)_ ability text. Alternatively, we could use 'that we specify'.

Yes, we mainly state the value on the unit's data sheet (in brackets, as part of the _Pulsar_ ability keyword, in the 'Notes' column...) — but in principle we might also state it elsewhere (in a paragraph that describes some unique special effect in a scenario, for example). However, I accept that this is very much a lesser consideration.

With some further tweaks, perhaps something like this?

```suggestion
. Determine the number of shots, based on the value of `(X)` that we specify for the ability on the unit's data sheet — for example, `{pulsar-d3plus1}` or `{pulsar-d6}`.
. Resolve that many {macro-weapon} shots against the target detachment.
```

modules/special-rules/pages/pulsar-x.adoc
digi-brain: > Ooh, That's a bit of a tough one. We wouldn't say 'Remember that Death Ray always represents...', so I think we should use 'Remember that a {pulsar}...'.

Haha, I wouldn't say that, but for a different reason. It depends on whether it is prose usage or 'keyword' usage, for want to a better description. (I often have to make this distinction in software documentation, and then adjust the text and formatting markup accordingly.)

- Prose usage: "Remember that a Death Ray weapon always represents..." — requires 'a' and 'weapon' and lacks special formatting for the term.
- Keyword usage: "Remember that _Death Ray_ always represents..." — omits 'a' and 'weapon' and uses the special formatting for the keyword.

I'll think about it some more. The hybrid usage of {pulsar} for both prose and keyword usage is already bugging me 😂 and maybe there's a neater way out of it.

modules/special-rules/pages/pulsar-x.adoc
digi-brain: > I don't remember who drafted that bit originally

Doesn't matter - we'll often think of new tweaks and ways to shape things when we're editing for one reason or another.

> but I'd prefer to keep it as an explicit list. Mainly because it gets forgotten in firefights.
>
>If anything, I'm inclined to make it more comprehensive by combining the old and new versions:

Fair enough. The 'or any other game purpose' bit covers my first concern — once I got on that track, I thought there was also an opportunity to make it more concise. But I'm happy to go with your suggestion here.

modules/special-rules/pages/pulsar-x.adoc
digi-brain: > P.S. I did a minimal edit as the existing text had already been checked.

Yeah, that's fine and is generally a good approach. I just often find that when I have new cause to edit something, then I'll spot and take other small opportunities to improve the text while I'm at it (or sometimes a new tweak creates new potential for ambiguity, so I seek to mitigate that).

modules/special-rules/pages/pulsar-x.adoc
IJW-Wartrader: I have no preference between 'we state' and 'we specify'.

But I think your first suggestion at https://github.com/GameBrains/er-core/pull/43#discussion_r1176229870 is more concise.

digi-brain: > But I think your first suggestion at [#43 (comment)](https://github.com/GameBrains/er-core/pull/43#discussion_r1176229870) is more concise.

Indeed it is — I just wasn't sure how strongly you felt about the "unit's data sheet" part of your last suggestion, but I'm happier to drop that compromise and be more concise.

I think that first suggestion would suffice — but arguably, part of my last suggestion improves on it slightly because it makes direct reference to the 'X' placeholder that exists in the topic/ability title (rather than the slightly indirect "value in brackets"):

```suggestion
. Determine the number of shots, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
. Resolve that many {macro-weapon} shots against the target detachment.
```
I think this is fine so I'll apply it shortly — you're always free to raise a new suggestion if you think that I've unwittingly created a new issue in this iteration that I didn't think of.

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
+= Pulsar (X)
+
+This weapon is unique to the Eldar race, and it demonstrates their mastery of laser weaponry.
+
+The weapon shoots in rapid but powerful bursts, guided by advanced targeting technologies.
+As a result, the weapon can destroy multiple targets in a single salvo.
+
+---
+
+When you shoot with a {pulsar-x}:
```
digi-brain, line 10: Hmmm... we previously figured out a consistent way to construct the lead-in sentence for super-heavy weapons, in part to overcome difficulties in relation to 'weapon systems' — but we broke it for _Pulsar_ because of the reference to 'shots'. If instead we apply the same construction here, we're more consistent and we solves the 'prose/not prose hybrid usage' problem we've been discussing at https://github.com/GameBrains/er-core/pull/43/files#r1177022647.

```suggestion
When you make a {pulsar-x} shot:
```

However, this makes the mention of 'shots' a little odd in the bullet that follows, so I think we'll need to tweak that a little more to work with this suggestion.

modules/special-rules/pages/pulsar-x.adoc
digi-brain: My suggestion at https://github.com/GameBrains/er-core/pull/43/files#r1186735306 will overcome part of this issue.

Then, for this note, I think we can follow through with something like this:

```suggestion
IMPORTANT: Although a {pulsar-x} shot results in a salvo of {macro-weapon} shots, it still only counts as a single super-heavy weapon.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+. Determine the number of shots, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
+. Resolve that many {macro-weapon} shots against the target detachment.
```
digi-brain, line 13: To tie in with my other suggestions, I think we need to reframe things here around the idea of a "salvo of {macro-weapon} shots". Perhaps something like:

```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.

. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
. Resolve that many {macro-weapon} shots against the target detachment.
```
...and then we can delete the note that we currently have below the bullets.

modules/special-rules/pages/pulsar-x.adoc
digi-brain: Scratch that — to tie in with my suggestion at https://github.com/GameBrains/er-core/pull/43/files#r1186738056, we can delete this line, and then modify the one after it.
```suggestion
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+IMPORTANT: Remember that a {pulsar-x} is a single super-heavy weapon.
+It does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```
digi-brain, line 16: To tie in with my other suggestions:
```suggestion
IMPORTANT: Remember, a salvo of {macro-weapon} shots from a {pulsar} weapon does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

These several separate but related suggestions I've made this evening are getting messy — so I think I'll apply them, so we can see how they really look together. Then we can discuss/tweak further as necessary.

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,20 @@
...
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
+. Resolve that many {macro-weapon} shots against the target detachment.
+
+IMPORTANT: Remember, {pulsar-x} represents a single super-heavy weapon -- the salvo of {macro-weapon} shots that result do *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```
digi-brain, line 17:
```suggestion
IMPORTANT: Remember, {pulsar-x} represents a single super-heavy weapon -- the salvo of {macro-weapon} shots that results does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,20 @@
...
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
+. Resolve that many {macro-weapon} shots against the target detachment.
+
+IMPORTANT: Remember, {pulsar-x} represents a single super-heavy weapon -- the salvo of {macro-weapon} shots that results does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```
digi-brain, line 17: Improve phrasing slightly:
```suggestion
IMPORTANT: Remember, {pulsar-x} represents a single super-heavy weapon -- the resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,20 @@
...
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
+. Resolve that many {macro-weapon} shots against the target detachment.
+
+IMPORTANT: Remember, {pulsar-x} represents a single super-heavy weapon -- the resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```
digi-brain, line 17: Rationalise the two IMPORTANT notes down to one:

```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
The resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.

. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
. Resolve that many {macro-weapon} shots against the target detachment.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+The resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
+. Resolve that many {macro-weapon} shots against the target detachment.
```
digi-brain, line 16: Insert footnote to highlight difference from _Epic 40,000_:

```suggestion
. Determine the number of shots in the salvo, based on the value of `(X)` that we specify — for example, _`(D3+1)`_ or _`(D6)`_.
footnote:[{e40k-diff}
The 'salvo' value of {pulsar} in {e40k} is D6 only.
This makes it hard to balance the performance and points costs of certain units.
So, we've added _`(X)`_ to support other values too, for more flexible design of units in the army lists.
]
. Resolve that many {macro-weapon} shots against the target detachment.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,24 @@
...
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+The resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify -- for example, _`(D3+1)`_ or _`(D6)`_.
+footnote:[{e40k-diff}
+The 'salvo' value of {pulsar} in {e40k} is D6 only.
```
digi-brain, line 17:
```suggestion
In {e40k}, the 'salvo' value of {pulsar} is D6 only.
```

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,24 @@
...
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+The resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
+
+. Determine the number of shots in the salvo, based on the value of `(X)` that we specify -- for example, _`(D3+1)`_ or _`(D6)`_.
+footnote:[{e40k-diff}
+In {e40k}, the 'salvo' value of {pulsar} is D6 only.
```
IJW-Wartrader, line 17:
```suggestion
In {e40k}, the 'salvo' value of {pulsar} is D6 only.
```

One of the reasons for updating the Pulsar is that it wasn't clear if it _was_ always D6 in Epic 40k. The Revenant's 'Pulse Laser' was formatted in exactly the same way as the Phantom's Pulsar, but with lower range and Firepower D3 instead of Firepower D6.

modules/special-rules/pages/pulsar-x.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+The weapon shoots in rapid but powerful bursts, guided by advanced targeting technologies.
+As a result, the weapon can destroy multiple targets in a single salvo.
+
+---
+
+When you make a {pulsar-x} shot:
+
+IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
+The resultant salvo of {macro-weapon} shots does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```
IJW-Wartrader, line 13: Shouldn't this be a single sentence? Or at least remove the duplication of 'results in a salvo of {macro-weapon} shots'/'The resultant salvo of {macro-weapon} shots'?

Something like this:
```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

Or:
```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but it still only counts as a single super-heavy weapon.
Therefore a {pulsar-x}  does *not* count as multiple super-heavy weapons for placement of Blast markers, suppression, contribution to firefights, or any other game purpose.
```

modules/special-rules/pages/pulsar-x.adoc
digi-brain: Hmm. I see your point, but I was simply trying to highlight the fact that the original _Pulsar_ rules text is explicitly D6 and nothing else, and we've made it more flexible.

Yes, there is the issue that it seems like the Pulse Laser on the Revenant Titan should work in the same way as the Pulsar, except with a different 'salvo' value, or else the Blast markers get silly — but I'm not sure we really need to spell all that out here, even if it is a major reason for the change.

_If we need to get into it at all, another option might be to mention it as another 'difference' note in close proximity to the Revenant Titan data sheet — the old one didn't mention the _Pulsar_ ability whereas the new one will state _Pulsar (D3)_ or whatever it ends up being._

If you think we **should** spell it out here then I'm not dead against it — feel free to suggest some text. You might want to resolve this thread and then suggest across multiple lines to integrate this specific point with the other more general points I made about balance, etcetera.

modules/special-rules/pages/pulsar-x.adoc
digi-brain: > Shouldn't this be a single sentence? Or at least remove the duplication of 'results in a salvo of {macro-weapon} shots'/'The resultant salvo of {macro-weapon} shots'?

Yes, I kinda ran out of steam after making several iterations in which I created, moved and recombined two separate notes, and needed to come back at another time with fresh eyes. 😁 It would indeed benefit from further rationalisation.

I think your first suggestion might work well, with some further small tweaks to both our contributions to it:

```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but we treat these collectively as only a single super-heavy weapon for all game purposes (such as placement of Blast markers, suppression and contribution to firefights).
```
It's a bit of a long sentence, but acceptable I think.
---

On a related thought: The 'salvo' effect of {pulsar-x} seems to duplicate the principles we outlined in the 'weapon systems' concept — IIRC it's just that the specific number of {macro-weapon} multiple shots in a {pulsar-x} salvo is variable rather than fixed as would be more usual for other weapon systems.

If so, then I wonder if it might make sense to integrate terminology between the two at some point, and/or make {pulsar-x} explicitly a 'weapon system' as opposed to a 'super-heavy weapon' per se?

(We don't need to do it now, it's just a thought to consider for the future.)

IJW-Wartrader: > I think your first suggestion might work well, with some further small tweaks to both our contributions to it:

The 'only' in the middle looks slightly odd. Minor suggestion:

```suggestion
IMPORTANT: A {pulsar-x} shot results in a salvo of {macro-weapon} shots -- but we treat these collectively as a single super-heavy weapon for all game purposes (such as placement of Blast markers, suppression and contribution to firefights).
```

----

> On a related thought: The 'salvo' effect of {pulsar-x} seems to duplicate the principles we outlined in the 'weapon systems' concept — IIRC it's just that the specific number of {macro-weapon} multiple shots in a {pulsar-x} salvo is variable rather than fixed as would be more usual for other weapon systems.

Ah crap. :-( 

I've just looked at our current definition for 'weapon systems' and it's out of date – multiple shots from a weapon system are **not** treated as a single super-heavy weapon!

Should I start a new branch for that, to prevent scope creep in this branch?

digi-brain:
@IJW-Wartrader For some reason, when I try to reply to your suggestion at https://github.com/GameBrains/er-core/pull/43#discussion_r1188389887, I can't for some reason. (The **Comment** button fades for a couple of seconds and then returns to normal, and that's it — it doesn't post my reply, and doesn't show an error message either!)

I intend to apply that suggestion anyway, and most of the reply relates the the new issue, so I'll paste it here instead:

---

> The 'only' in the middle looks slightly odd. Minor suggestion:

Ah yes, I think you're right — I'll apply this latest iteration shortly. Once again, two heads/multiple iterations prove to be better than one! 😂

---

> Ah crap. :-(
> I've just looked at our current definition for 'weapon systems' and it's out of date – multiple shots from a weapon system are not treated as a single super-heavy weapon!

Oh right — I have a vague memory that you mentioned something about that a while back. I guess your play-testing revealed that 'treat as a single SHW for Blast markers etcetera' didn't work well in practice for some reason — something to the effect that while the likes of an Inferno Gun with `N x {macro-weapon}` shots **seems** like it'd be overkill on Blast markers, it actually isn't such a big deal in practice and the way we have it now is actually underpowered?

If memory serves me correctly, this will be a reversal to one of the major reasons why we created the 'weapon systems' concept in the first place — but the other reasons still stand, such that 'weapon systems' is still a useful concept? (Ref: Current published instance of [Game effects of weapon systems](https://epicremastered.com/core/special-rules/shooting-with-war-engines.html#_game_effects_of_weapon_systems))

Better raise a new issue then I think, and you can outline as appropriate there.

> Should I start a new branch for that, to prevent scope creep in this branch?

Yes indeed, that'd be best — and we can link the new issue ticket to it. The 'weapon systems' stuff was relatively complex to edit to get the Blast marker bit right (as we intended it at the time, that is), but I think there's a good chance that the re-edit to reverse this aspect will be somewhat easier. All the same, it's generally best to favour separate branches with narrow scopes.

digi-brain:
I'll build locally soon and do final checks with a view to publish this iteration, barring last-minute surprises.

digi-brain (approved):
Completed final local checks — all good, ready to merge.

